PR #9, opened 2020-11-04
pieces edits

localghost443:


docs/pieces.md
```diff
@@ -2,21 +2,16 @@
...
 
-# Pieces in A3
+# Pieces
+What is a Piece?
 
-:::  tip Note:
-<!-- I realize that for alpha 1 our audience is more familiar with Apostrophe, but this is a topic that comes up even among people who know a lot about Apostrophe, so I'd like this note to stay in. So if you have a
-concern with it, work with me to make it better, don't remove it. -Tom -->
-**"What's a piece?"**
+As experienced A2 delevelopers know, a piece type is a repository for a collection of related content-types or pieces. If you have hundreds of documents of the same type, that's a piece. If you want to let the user view them in a certain sorting order, that's a piece. If the same item might be appropriate to display on many different pages, that's a piece. Blog posts, events and products are all excellent examples. Less obvious examples include "category" or "tag" content types that serve to create a taxonomy for other pieces.
```
abea, line 8: typo: "delevelopers"

> a piece type is a repository for a collection of related content-types or pieces

That is less clear to me. "piece types define structure and behavior for content categories, such as blog posts, products, or events."

> If you have hundreds of documents of the same type, that's a piece.
That's not enough in and of itself, which is why combining the criteria in that sentence is useful.

It looks like the question is on the first sentence. My suggestion: "As experienced A2 developers know, piece types define structure and behavior for categories of content, such as blog posts, products, or events." We don't need to go for a full Websters Dictionary definition.

docs/pieces.md
```diff
@@ -202,21 +197,21 @@ In A2, methods were added in `construct`, which no longer exists. In A3, your cu
...
-Sometimes we want to extend an existing method we inherited from our base class, while still taking advantage of it. In A2 we did that "by hand," saving the old method in a variable. This was not intuitive for new developers.
+Sometimes we want to extend an existing method we inherited from our base class. In A2 we did that by saving the old method in a variable. This was not intuitive for new developers.
 
-A3 modules replace this with the `extendMethods` section, which provides a clear pattern. In the example above, the `generate` method is extended to generate a random `price`. Notice that `_super` is always the first argument here. Invoking it calls the old version of the method, as part of our implementation of the new one.
+A3 modules replace this process with the `extendMethods` section, providing a clear pattern. In the example above, the `generate` method is extended to generate a random `price`. Notice that `_super` is always the first argument. Invoking it calls the old version of the method, as part of our implementation of the new one.
 
 This allows us to run:
 
-```bash
+```sh
```
abea, line 206: This isn't consistent, but also not critical.

docs/pieces.md
```diff
@@ -327,13 +322,13 @@ These are the important sub-sections of a query builder:
 
 ## REST APIs for pieces
 
-REST APIs are automatically made available for piece types. For more information, see [REST APIs](rest-apis.md).
+REST APIs are automatically made available for piece-types. For more information, see [REST APIs](rest-apis.md).
```
abea, line 325: I don't think this should be hyphenated. `piece-type` is a module name. "Piece type" is a concept.

docs/pieces.md
```diff
@@ -346,7 +341,7 @@ modules: {
 }
 ```
 
-```
+```js
```
abea, line 344: Isn't it `javascript`?

abea (changes requested):

docs/pieces.md
boutell: I agree with Bea's points here. Also we need to avoid confusing people about a key point: a piece type module defines ONE content type. And, we can't lose the distinction from pages.

How about:

"As experienced A2 developers know, a piece type module defines structure and behavior for a single type of content, such as blog posts, products or events, that shouldn't be represented as a page in a tree."

And the rest of the paragraph would stay as-is, it sounds like.

docs/pieces.md
boutell: Can we please go back to `bash`, we use bash-isms sometimes in our command lines. None of us use sh and none of us, to my knowledge, know what the forbidden "bash-isms" are (but `&&`, which we use like candy, is one of them).

docs/pieces.md
```diff
@@ -202,21 +197,21 @@ In A2, methods were added in `construct`, which no longer exists. In A3, your cu
...
+```sh
 node app product:generate
 ```
 
 To generate test products with prices as well as titles.
 
 ### Handlers
 
-In A2, pieces had initially empty methods like `beforeInsert` and `afterUpdate` that were invoked as part of those operations. This was useful, but could be bug-prone when piece types extended each other.
+In A2, pieces initially had empty methods like `beforeInsert` and `afterUpdate` that were invoked as part of those operations. This was useful, but could be bug-prone when piece types extended each other.
```
boutell, line 214: The original wording is correct. These still exist in A2. They are initially empty, not initially existing (and no longer existing).

docs/pieces.md
boutell: Correct, no hyphen.

docs/pieces.md
```diff
@@ -327,13 +322,13 @@ These are the important sub-sections of a query builder:
 
 ## REST APIs for pieces
 
-REST APIs are automatically made available for piece types. For more information, see [REST APIs](rest-apis.md).
+REST APIs are automatically made available for piece-types. For more information, see [REST APIs](rest-apis.md).
 
 ## Pieces and widgets
 
-Of course, it's helpful to be able to display pieces anywhere on the site. Sometimes we don't want a browseable piece page at all, sometimes we want to "tease" that piece as the "featured product" on the home page, and so on.
+Of course, it's helpful to be able to display pieces anywhere in your project. Sometimes you won't want a browseable piece page at all, sometimes or you may want to "tease" that piece as the "featured product" on the home page, and so on.
```
boutell, line 329: Grammar: "sometimes or you"

docs/pieces.md
boutell: Both are valid and js is more succinct so I'm on board.

docs/pieces.md
```diff
@@ -380,4 +375,4 @@ Pieces often implement async components as an easy way to display them on the pa
 
 ## Piece pages
 
-As referred to above, while they are not needed in every case, [piece pages](piece-pages.md) are a crucial companion to pieces for any website that allows pieces to be browsed by the user. See the [piece pages](piece-pages.md) section to get started.
+As referred to above, while they are not needed in every case, [piece pages](piece-pages.md) are a crucial companion to pieces for any project that allows the user to browse an index of pieces. See the [piece pages](piece-pages.md) section to get started.
```
boutell, line 378: `website` is correct here. piece pages are not typically relevant in headless projects. The developer would likely just use the REST API of their piece module.

boutell (changes requested):

docs/pieces.md
abea: I was trying to avoid "type of content" or "content type" since it seems close to defining a word with itself. So I'd change my bit to "a piece type defines the structure and behavior for a category of content"

abea: The page tree thing seems irrelevant. It's true, but between examples below and devs' experience, I don't think it needs to be stated here.

docs/pieces.md
abea: That works.

docs/pieces.md
boutell: Hmm I guess it would be hard to confuse a page with anything suited to these use cases, maybe it doesn't need to be in sentence one after all. But I'd like the distinction to at least be mentioned nearby.

boutell: Drupal, Wordpress and other CMSes use "content type" when explaining similar concepts. I think it's clear and concise.

docs/pieces.md
localghost443: "initialized with" empty methods? The wording is still confusing to me.

docs/pieces.md
abea: "Content type" is what Drupal calls the thing. Wordpress uses "post type." We have "piece type," so I want to avoid "type" in the definition in order to be a sort of definition. But not worth bike shedding now. I'll make an update for review.

abea (approved):

boutell (approved):

PR #10, opened 2020-11-04
Bea edits alpha 1

abea:


localghost443 (approved):

docs/module-format-example.md
```diff
@@ -6,7 +6,7 @@ title: "Module Format Example"
 
 The new A3 module format is best understood by reviewing a complete example with all of the sections.
 
-> As experienced A2 developers will know, this is "what goes in `index.js`."
+As experienced A2 developers will know, this is "what goes in `index.js`."
```
localghost443, line 9: I would like to avoid using sentence fragments in quotes as if we're quoting someone. We should explain this directly.

abea: true

abea: I think he's referencing how we'll often refer to this. I'll add clarification.

boutell (approved):

PR #24, opened 2021-01-12
API documentation for mode, locale, and publishing

abea:
This documentation covers the mode and locale query parameters for API routes as well as the `/publish` routes for pieces and pages.


docs/.vuepress/styles/index.styl
```diff
@@ -0,0 +1,3 @@
+table code {
+  white-space: nowrap;
```
abea, line 2: The routes wrapping in markdown tables were extra hard to read.

docs/guide/rest-apis.md
```diff
@@ -141,7 +141,7 @@ Note `pages` and `currentPage`. For performance reasons, not all of the products
 GET /api/v1/product?page=2
 ```
 
-### `GET /api/v1/product/:docId`
```
abea, line 144: We now have a document property that looks like `docId` (`aposDocId`), so I changed this to be less confusing.

docs/guide/rest-apis.md
```diff
@@ -297,25 +297,33 @@ Returns an individual page as a JSON object:
...
+### `PATCH /api/v1/@apostrophecms/page/:id`
 
 **Authentication required.** Identical to PATCHing a piece, except that you may include `_position` and `_targetId` if you wish to change the position in the tree. To easily move a page into the trash, set `_targetId` to `_trash` and `_position` to `lastChild`. You may similarly move pages out of the trash by moving them to a position relative to a page that is not in the trash.
 
-### `DELETE /api/v1/@apostrophecms/page:docId`
+### `DELETE /api/v1/@apostrophecms/page/:id`
 
 **NOT supported.** Instead set `_targetId` to `_trash` and `_position` to `lastChild`. You may similarly move pages out of the trash by moving them to a position relative to a page that is not in the trash.
 
-## PATCHing a document of any type
```
abea, line 312: Not possible anymore.

docs/guide/rest-apis.md
```diff
@@ -297,25 +297,33 @@ Returns an individual page as a JSON object:
...
+## Locale and mode in API calls
```
abea, line 312: There's detail here, especially in the sub-section, that is helpful for routes but is too much to put in the reference pages.

docs/reference/api/pages.md
```diff
@@ -13,6 +15,18 @@ All [page types](TODO) use a single set of API endpoints, unlike piece types. Di
...
 |DELETE | Not supported | [See more in PATCH request details.](#moving-pages-to-the-trash) | n/a |
 
+### Additional page endpoints
+
+| Method | Path | Description | Auth required |
+|---------|---------|---------|---------|
+|POST | [`/api/v1/@apostrophecms/page/:id/publish`](#post-api-v1-apostrophecms-page-id-publish) | Publish the draft version of a page | TRUE |
+
+<!-- TODOcument -->
+<!-- |GET | [`/api/v1/@apostrophecms/page/:id?locales=1`](#get-api-v1-apostrophecms-page-id-locales-1) | Request the available locales for a specific page | TRUE |
```
abea, line 25: We're not writing these up yet, but it took a while to work all this out so I thought I'd leave it for later.

docs/reference/api/pages.md
```diff
@@ -182,11 +198,20 @@ Individual page objects will include `_children` and `_ancestor` arrays, as well
...
+|`apos-locale` | `?apos-locale=fr` | Set to [a valid locale](#TODO) to request the page document version for that locale. |
+
+Read more about [mode and locale parameters on single-document requests](/guide/rest-apis.md#locale-and-mode-in-single-document-requests).
+
 ### Request example
 
 ```javascript
 // Request inside an async function.
-const response = await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr?apikey=myapikey', {
+const response = await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr:en:published?apikey=myapikey', {
```
abea, line 214: Ids changed.

docs/reference/glossary.md
```diff
@@ -0,0 +1,7 @@
+# Glossary
+
+<!-- This is a work in progress. It is not added to the navigation yet, but exists to begin documenting terms. -->
+
+## Document set
```
abea, line 5: We need some word for it. We can change this later if we have something better, at which point we can also replace it in the text.

localghost443 (approved):

docs/guide/rest-apis.md
boutell: `_id` would be exactly correct and then they'd know what it is for sure.

docs/guide/rest-apis.md
```diff
@@ -190,7 +190,7 @@ In addition, we extend the REST convention with MongoDB-style operators:
 }
 ```
 
-### `DELETE /api/v1/product:docId`
+### `DELETE /api/v1/product/:id`
 
 **NOT supported. Instead PATCH the `trash` property to `true`.** Apostrophe never discards your work, so the `DELETE` verb is not applicable. However you may mark any piece as trash by `PATCH`ing that property. You may bring it back later by setting the property to `false`.
```
boutell, line 195: Update: this is now supported. However if you do it in draft mode you should be aware that the published document will also be deleted, and that this is not the way to "discard a draft." You should also be aware that Apostrophe's general UI philosophy is not to delete things outright, except for drafts that have never been published, so PATCHing the `trash` property is often still the best policy.

docs/guide/rest-apis.md
```diff
@@ -297,25 +297,33 @@ Returns an individual page as a JSON object:
...
 
-## PATCHing a document of any type
+## Locale and mode in API calls
 
-**Authentication required.** As a convenience, you may also `PATCH` a document without specifying the exact module that should handle it:
+Once published, most Apostrophe document types have both a draft and a published copy in the database. If you are using multiple locales in your project, there may be documents for each locale as well. The `apos-mode` and `apos-locale` query parameters help support using REST API routes to manage those versions.
 
-### `PATCH /api/v1/@apostrophecms/doc/:docId`
+Passing `draft` or `published` values on `apos-mode` and your locale names on the `apos-locale` parameter tell Apostrophe to execute that request for those versions of the piece or page you are working with. For example, `GET /api/v1/product?apos-mode=draft&apos-locale=fr` will get all the draft and `fr` (French) locale versions of the product documents. These can be used independently, so you can only use `apos-mode` and Apostrophe will use the default locale value. Or only use `apos-locale` and Apostrophe will use `published` for the mode.
```
boutell, line 316: It should be mentioned that locale support is only partially complete in alpha 3, while mode support is 100% complete.

docs/guide/rest-apis.md
```diff
@@ -297,25 +297,33 @@ Returns an individual page as a JSON object:
...
 
-In this case the request is routed to the appropriate module. Apostrophe uses this route for in-context editing.
+### Locale and mode in single-document requests
+
+**The exception to those assumed values are with `GET ONE`, `PATCH`, and other requests that include a document `_id` value.** Document IDs include mode and locale information already. `ckgsj5in400d5xi4lb6fu29rh:fr:draft` ends with `:fr:draft`, indicating that it is the `fr` locale and `draft` version of a particular piece or page.
```
boutell, line 320: I don't love `GET ONE` but I don't see another way to make that distinction, so ✅

docs/guide/rest-apis.md
```diff
@@ -297,25 +297,33 @@ Returns an individual page as a JSON object:
...
+
+You may still use the `apos-mode` and `apos-locale` parameters with these requests related to existing documents. If the values for those parameters match the document you are requesting, the request will treat it as if the parameters weren't there. If the values differ from the `_id` value, and a version of the document exists for the mode and locale values, the request will apply to the version matching the parameter values.
```
boutell, line 322: I would just say: "However, if `apos-mode` and/or `apos-locale` query parameters are present, they always take precedence. This makes it easy to request an alternate version of a document whose `_id` you already have."

To say more can lead to inaccuracies. For instance, it is not true that it will only do this if the other version exists. It will definitely do it (and you will get a 404, if appropriate, or create a new document, etc).

docs/guide/rest-apis.md
```diff
@@ -297,25 +297,33 @@ Returns an individual page as a JSON object:
...
+
+For example, a `GET` request to `/api/v1/product/ckgsj5in400d5xi4lb6fu29rh:fr:published` will normally return a product document in the `fr` locale and that is published.
+
+If we add an `apos-mode` parameter using the same route, we can get the draft version. So a `GET` request to `/api/v1/product/ckgsj5in400d5xi4lb6fu29rh:fr:published?apos-mode=draft` returns the draft version, even though the route includes the published mode. This allows developers to work on multiple versions of a single document without repeatedly updating the `_id` value in their code.
```
boutell, line 326: This example is great.

docs/reference/api/pages.md
```diff
@@ -13,6 +15,18 @@ All [page types](TODO) use a single set of API endpoints, unlike piece types. Di
...
+| Method | Path | Description | Auth required |
+|---------|---------|---------|---------|
+|POST | [`/api/v1/@apostrophecms/page/:id/publish`](#post-api-v1-apostrophecms-page-id-publish) | Publish the draft version of a page | TRUE |
+
+<!-- TODOcument -->
+<!-- |GET | [`/api/v1/@apostrophecms/page/:id?locales=1`](#get-api-v1-apostrophecms-page-id-locales-1) | Request the available locales for a specific page | TRUE |
+|POST | [`/api/v1/@apostrophecms/page/:id/export`](#post-api-v1-apostrophecms-page-id-export) | Copy a page from one locale to another | TRUE |
+|POST | [`/api/v1/@apostrophecms/page/:id/revert-draft-to-published`](#post-api-v1-apostrophecms-page-id-revert-draft-to-published) | Revert a draft page to the the state of the published version, if available | TRUE |
+|POST | [`/api/v1/@apostrophecms/page/:id/revert-published-to-previous`](#post-api-v1-apostrophecms-page-id-revert-published-to-previous) | Revert a published page to the previous version, if available | TRUE | -->
+
```
boutell, line 29: Since we are covering `revert-published-to-previous` we should also cover the new `unpublish` route, which deletes the published mode while leaving the draft mode in place for that document. It may not be used if any descendant pages are still published.

boutell (changes requested):

boutell (changes requested):
This is good stuff, hopefully simple changes in the notes of the review.

docs/guide/rest-apis.md
abea: I thought I'd seen that used elsewhere, but maybe not. I'm open to changing this.

docs/guide/rest-apis.md
abea: This is correct. In the interest of getting this out and keeping this focused on the locale and mode stuff I'd like to get to that in a future docs update. It's not likely scaring anyone off at the moment.

abea: I've created an internal ticket for this.

boutell (approved):

PR #30, opened 2021-02-08
Field types 2

abea:
Moar field type pages


docs/reference/field-types/checkbox.md
```diff
@@ -0,0 +1,66 @@
+# `checkboxes`
```
localghost443, line 1: @abea @boutell Is this our lonely plural guy we didn't make singular?

docs/reference/field-types/checkbox.md
```diff
@@ -0,0 +1,66 @@
...
+    {
+      label: 'Romance 🥰',
+      value: 'romance'
+    },
+    {
+      label: 'Comedy 🤣',
+      value: 'comedy'
+    },
+    {
+      label: 'Crime 😰',
```
localghost443, line 22: 🕵️

localghost443 (approved):

docs/reference/field-types/checkbox.md
abea: Would have been better.

abea: Accepting PRs.

docs/reference/field-types/checkbox.md
abea: The fields weren't singularized like modules, if that's what you're thinking of. This might be an odd duck, but it's not because of that.

boutell: This is plural because it is *always* an array, which is a distinction worth making, although its singular equivalent happens to be named `boolean`. (This one really ought to be called `booleans`! We don't usually name our fields after UI choices for representing them. But that's not something to hold this PR for.)

docs/reference/field-types/checkbox.md
boutell: ❤️ this whole thing.

boutell (approved):

PR #32, opened 2021-02-11
Fragments

abea:
It's still written in the migratory perspective, but hopefully in such a way as to be easily updated for new devs later. Ultimately we'll separate "widgets and templates," but that was out of scope for this mission.

### DIFF
--- a/docs/guide/widgets-and-templates/fragments.md
+++ b/docs/guide/widgets-and-templates/fragments.md
@@ -28,10 +28,10 @@ One difference from macros is that they are then called using the `render` templ
 
 ```django
 {% render heading() %}
-{# `<h2>This is a heading fragment</h2>` #}
+{# Renders: `<h2>This is a heading fragment</h2>` #}
 ```
 
-You can also pass arguments to macros, which is key to using them across templates. Those arguments can be any data or variables understood by Apostrophe templates (e.g., strings, objects, arrays, etc.).
+You can also pass arguments to fragments, which is key to using them across templates. Those arguments can be any data or variables understood by Apostrophe templates (e.g., strings, objects, arrays, etc.).
 
 ```django
 {% fragment heading(adjective) %}
--- a/docs/guide/widgets-and-templates/fragments.md
+++ b/docs/guide/widgets-and-templates/fragments.md
@@ -1,0 +1,148 @@
+# Fragments are the new macros
+
+[Nunjucks macros](https://mozilla.github.io/nunjucks/templating.html#macro) are great tools for sharing template code across files. They have one significant limitation, however, in that they run synchronously. If you include template tags in them that run asynchronously, those tags will seem to not run at all when it comes to rendering.
+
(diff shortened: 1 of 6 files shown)


docs/guide/widgets-and-templates/fragments.md
```diff
@@ -0,0 +1,148 @@
...
+{% fragment heading() %}
+  <h2>This is a heading fragment</h2>
+{% endfragment %}
+```
+
+One difference from macros is that they are then called using the `render` template tag:
+
+```django
+{% render heading() %}
+{# `<h2>This is a heading fragment</h2>` #}
```
boutell, line 31: I'm not sure the comment syntax is a good way to express this, maybe a separate markup code block showing the output.

docs/guide/widgets-and-templates/fragments.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+You can also pass arguments to macros, which is key to using them across templates. Those arguments can be any data or variables understood by Apostrophe templates (e.g., strings, objects, arrays, etc.).
```
boutell, line 34: To fragments.

docs/guide/widgets-and-templates/fragments.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+## Fragment limitations
+
+In comparing fragments to macros, it is worth noting a couple of macro features that are not *yet* supported in fragments.
+
+
+### Default parameter values
+
+Macros can include default values for their arguments. This is not currently available in fragments. As a work-around, you can set variables in fragments that are overwritten if the argument is provided.
```
boutell, line 119: Nice

boutell (changes requested):

docs/guide/widgets-and-templates/fragments.md
abea: Improved to `{# Renders: `<h2>This is a heading fragment</h2>` #}`?

docs/guide/widgets-and-templates/home-page.md
```diff
@@ -0,0 +1,16 @@
...
+// modules/@apostrophecms/home-page/views/page.html
+{% extends "layout.html" %}
+
+{% block main %}
+  {% area data.page, 'main' %}
+{% endblock %}
+```
+
+- Areas are added to the template with the new `area` nunjucks tag. There is no `apos.area` helper function anymore.
+- You no longer configure the area here. This is done exclusively in the `index.js` file for the page-type or piece-type. In the template, just pass the page and the name of the area.
```
boutell, line 15: Which means this example is not sufficient, so an example index.js should to be given too, declaring the main area as a field.

boutell (changes requested):
Just one change, re: home pages.

docs/guide/widgets-and-templates/home-page.md
abea: This was transferred from the `widgets-and-templates.md` file verbatim. The whole page needs to be rewritten or replaced, but I don't think this is the place to do it. Configuring areas is covered elsewhere as well.

boutell (approved):

PR #33, opened 2021-02-12
Field types 4

abea:


docs/reference/field-types/relationship-reverse.md
```diff
@@ -0,0 +1,35 @@
...
+_pizzas: {
+  type: 'relationshipReverse',
+  withType: 'pizza',
+  reverseOf: '_fabrics'
+}
+```
+
+We can now see `_pizzas` as a property of each `topping` object that is related to a product.
+
+If desired, we can specify `schema` and `relationshipsField` just as we would for `joinByArray`. Currently these are not automatic in a reverse join and must be fully specified if relationship properties are to be accessed. Most array joins do not have relationship properties and thus do not require reverse access to them.
```
boutell, line 17: `joinByArray` is no longer a thing. Also if this isn't automatic that's a bummer that should 100% get fixed, it makes me wonder if we should wait on documenting relationship-reverse until it is fixed. Because there's no possible valid value for these properties, other than what is already set on the relationship pointing back at us.

docs/reference/field-types/string.md
```diff
@@ -0,0 +1,71 @@
...
+:::
+
+## Use in templates
+
+The Nunjucks [nl2br](https://mozilla.github.io/nunjucks/templating.html#nl2br) tag can help print textarea strings with line breaks.
+
+```django
+<h2>{{ data.piece.dogName }}</h2>
+<p>
+  {{ data.piece.biography | nl2br | safe }}
```
boutell, line 69: Actually, this is a bad idea. If there are tags present in `biography` this will render them as HTML, creating CSRF risks or just unexpected behavior. Are you sure `safe` is necessary to make `nl2br` work?

boutell (changes requested):

docs/reference/field-types/relationship-reverse.md
abea: > makes me wonder if we should wait on documenting relationship-reverse until it is fixed.

I wasn't clear about this, but that's what I did, which is why `joinByArray` is still there. I brought the page over and started in on it, but I couldn't test it successfully so I didn't update this page. You can see it commented out in the sidebar nav file so it's not findable. There's a ticket, A30U-983, to dig into the field type.

abea: Note added. I commented out the only link to this page.

docs/reference/field-types/string.md
abea: Nunjucks recommends ` | striptags(true) | escape | nl2br`, it turns out. That strips out other tags before doing the `<br>` bit.

boutell: Verbose but it makes sense.

boutell (approved):

docs/reference/field-types/relationship-reverse.md
```diff
@@ -0,0 +1,36 @@
+# `relationshipReverse`
+
+<!-- TODO: This whole page needs to be revisited following successful testing of the `relationshipReverse` field type. -->
```
localghost443, line 3: :)

docs/reference/field-types/string.md
```diff
@@ -0,0 +1,71 @@
...
+<!-- TODO: 2.x options not yet available -->
+<!-- |contextual | Boolean | false | If `true`, it will prevent the field from appearing in the editor modal | -->
+<!-- |pattern | String | | Regular expression to validate entries |
+|patternErrorMessage | String | | Error message to display if `pattern` does not match | -->
+<!-- |searchable | Boolean | true | If false, content from the area will not appear in search results. | -->
+<!-- |readOnly | Boolean | false | If `true`, prevents the user from editing the field value | -->
+
+## `sortify`
+
+Setting `sortify: true` creates a parallel version of the field that is more intuitive for sorting purposes. This new, additional property's key will be the string field's name, appended with `Sortified`. Its value will be fully lowercase and have all punctuation removed. Apostrophe will automatically use it if a request is made to sort on the original field.
```
localghost443, line 54: should it be `key`

localghost443: should `Sortified` be capitalized?

docs/reference/field-types/string.md
```diff
@@ -0,0 +1,71 @@
...
+
+For instance, if your field's `name` is `lastName` and you set `sortify: true`, `lastNameSortified` will automatically be created and used when sorting on the `lastName` field. This provides case-insensitive sorting that also ignores punctuation differences.
+
+::: tip NOTE
+If you add `sortify: true` to an existing field, existing objects will get the sortified version of the field on the next deployment via the `apostrophe-migrations:migrate` command line task, when the individual Apostrophe documents are saved, or at the next startup when in development on your computer. Migrations like this only need to be run once because on future updates or inserts of a document the sortified property is automatically set.
```
localghost443, line 59: This could possibly be a list? Idk. There seems to be 3 time this is invoked and its difficult to read them as a long sentence. In the leave, "development *on your computer*" seems redundant

docs/reference/field-types/url.md
```diff
@@ -0,0 +1,43 @@
+# `url`
+
+`url` adds an editable URL field to the schema.
+
+Apostrophe will detect common mistakes, including leaving off `https://`. Common XSS attack vectors are laundered and discarded. Only "safe" URL schemes, e.g., `http`, `https`, `ftp` and `mailto`, are permitted.
```
localghost443, line 5: "Common XSS attack vectors are laundered and discarded. Only "safe" URL schemes, e.g., `http`, `https`, `ftp` and `mailto`, are permitted."

localghost443 (approved):

localghost443:
Approved but left some comments

docs/reference/field-types/string.md
abea: Yeah, that's the actual string that's appended.

abea: key isn't code you're going to write, so I didn't code-ify it.

docs/reference/field-types/string.md
abea: I'm not sure what you are saying should be in a list.

PR #37, opened 2021-02-24
Documents the DELETE REST API routes

abea:
Resolves https://apostrophecms.atlassian.net/browse/A3DOCS-35, updating DELETE API route options

### DIFF
--- a/docs/reference/api/authentication.md
+++ b/docs/reference/api/authentication.md
@@ -75,7 +75,7 @@ In case of an error an appropriate HTTP status code is returned.
 
 ### Usage
 
-Add this token to all REST API requests that require login by passing an `Authorization` header. This includes `POST`, `PUT`, `PATCH` and `DELETE` operations. You will also get more complete results from `GET` operations, including access to documents with `visibility` set to `loginRequired`.
+Add this token to all REST API requests that require login by passing an `Authorization` header. This includes `POST`, `PUT`, `PATCH` and `DELETE` operations. Using authorization for `GET` operations will return more complete results, including access to documents with `visibility` set to `loginRequired`.
 
 Your `authorization` HTTP header should look like:
 
--- a/docs/reference/api/authentication.md
+++ b/docs/reference/api/authentication.md
@@ -45,7 +45,7 @@ http://example.net/api/v1/article?apikey=myapikey1029384756
 
 ## Bearer tokens
 
-Bearer tokens are more appropriate for browser use and headless applications because they are tied to a single user account. To obtain a bearer token, make a POST request to `/api/v1/login/login`, with the `username` and `password` body properties:
+Bearer tokens are more appropriate for browser use and headless applications because they are tied to a single user account. To obtain a bearer token, make a `POST` request to `/api/v1/login/login`, with the `username` and `password` body properties:
 
 ```javascript
 // Request inside an async function.
@@ -75,7 +75,7 @@ On error an appropriate HTTP status code is returned.
 
 ### Usage
 
-Add this token to all REST API requests that require login by passing an `Authorization` header. This includes POST, PUT, PATCH and DELETE operations. You will also get more complete results from GET operations, including access to documents with `visibility` set to `loginRequired`.
+Add this token to all REST API requests that require login by passing an `Authorization` header. This includes `POST`, `PUT`, `PATCH` and `DELETE` operations. You will also get more complete results from `GET` operations, including access to documents with `visibility` set to `loginRequired`.
 
 Your `authorization` HTTP header should look like:
 
@@ -85,12 +85,12 @@ Bearer random123Token456xyz
 
 Note the need for the word "Bearer" before the key.
 
-To log out and destroy the token, send a POST request to `/api/v1/login/logout`, with the same `authorization` header. No body properties are required. After logging out, the token is no longer accepted.
+To log out and destroy the token, send a `POST` request to `/api/v1/login/logout`, with the same `authorization` header. No body properties are required. After logging out, the token is no longer accepted.
 `
 
 ## Session cookies
 
-As an alternative to a bearer token, you may request a session cookie. This is the mechanism Apostrophe's admin user interface uses to log in. Session cookies will automatically persist across tabs, but there is slightly more overhead to each request. To use this method, include `session: true` in the POST request to `/api/v1/@apostrophecms/login/login`.
+As an alternative to a bearer token, you may request a session cookie. This is the mechanism Apostrophe's admin user interface uses to log in. Session cookies will automatically persist across tabs, but there is slightly more overhead to each request. To use this method, include `session: true` in the `POST` request to `/api/v1/@apostrophecms/login/login`.
 
 ```javascript
 // Request inside an async function.
@@ -113,7 +113,7 @@ A successful response will return a session cookie via the `Set-Cookie` header, 
 
 ### End session
 
-Using the session cookie, send a POST request to `/api/v1/@apostrophecms/login/logout` to end the user session.
+Using the session cookie, send a `POST` request to `/api/v1/@apostrophecms/login/logout` to end the user session.
 
 ```javascript
 // Request inside an async function.
--- a/docs/reference/api/authentication.md
+++ b/docs/reference/api/authentication.md
@@ -71,7 +71,7 @@ A successful response will return a JSON object with a `token` property.
 }
 ```
 
-On error an appropriate HTTP status code is returned.
+In case of an error an appropriate HTTP status code is returned.
 
 ### Usage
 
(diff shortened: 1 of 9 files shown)


docs/reference/api/pages.md
```diff
@@ -373,6 +373,47 @@ The successful PATCH request returns the complete patched document. See the [pag
...
+await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr:en:published?apikey=myapikey', {
+  method: 'DELETE'
+});
+```
+
+### Response
+
+The successful DELETE request simply responds with a `200` HTTP response status code. On error an appropriate HTTP status code is returned. If the error is due to one of the rejection cases documented above, a message will be included to that effect, such as:
+
+```
```
abea, line 409: I'm including this for pages since there are specific rejection cases. I don't think the same is true for pieces.

boutell (changes requested):
This is great. In addition, it does not make sense to delete the draft if the published mode still exists. Please document that fact. I am opening a ticket to correctly throw an error in that situation. And, uh... I am also opening a ticket to check permissions on delete. 😆 Which doesn't matter much yet but will as soon as our full permissions model is in.

abea:
Okay, since you're doing that I'll document that it's a rejection case for both pages and pieces.

boutell (approved):

docs/reference/api/pages.md
```diff
@@ -373,6 +373,48 @@ The successful PATCH request returns the complete patched document. See the [pag
 
 The trash is part of the overall page tree in order to maintain the nesting structure. As such, there is not only a simple `trash` property to set `true`. Instead, set `_targetId` to `_trash` and `_position` to `lastChild` (or another position within the trash). You may similarly move pages out of the trash by moving them to a position relative to another page that is not in the trash.
 
+## `DELETE /api/v1/@apostrophecms/page/:_id`
+
+
+**Authentication required.**
+
+This API route **permanently deletes the page database document**. Moving pieces to the trash in the Apostrophe user interface or [using a PATCH request](#moving-pages-to-the-trash) do not permanently delete database documents and are typically better options.
```
localghost443, line 381: I don't think we need 'This API route'. Could leave it at 'Permanently' etc. I assume we're in an API route section already. I would also remove the part about 'typically better options'. We don't know that, we can trust their intent. I do think its useful to link to PATCH requests as you have.

docs/reference/api/pages.md
```diff
@@ -373,6 +373,48 @@ The successful PATCH request returns the complete patched document. See the [pag
...
+
+DELETE requests will be rejected if:
+- the `_id` matches the draft mode of a page that has an existing published mode document, or
```
localghost443, line 384: Maybe we don't need 'or' if this is a list. Unless there is an 'and' hidden somewhere

docs/reference/api/pages.md
```diff
@@ -373,6 +373,48 @@ The successful PATCH request returns the complete patched document. See the [pag
...
+- the `_id` matches the home page (`slug: '/'`), or
+- the `_id` matches a page that has sub-pages, or "children," in the page tree structure. You must delete child pages first.
```
localghost443, line 386: Should "children," have the comma inside, or at all?

docs/reference/api/pages.md
```diff
@@ -373,6 +373,48 @@ The successful PATCH request returns the complete patched document. See the [pag
...
+```javascript
+// Request inside an async function.
+await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr:en:published?apikey=myapikey', {
+  method: 'DELETE'
+});
+```
+
+### Response
+
+The successful DELETE request simply responds with a `200` HTTP response status code. On error an appropriate HTTP status code is returned. If the error is due to one of the rejection cases documented above, a message will be included to that effect, such as:
```
localghost443, line 408: Should DELETE be inline code here or else have *some* special treatment? "On error" is strange phrasing that I can't put my finger on. It's probably fine. "When in error", "If an error occurs", etc.

docs/reference/api/pages.md
```diff
@@ -373,6 +373,48 @@ The successful PATCH request returns the complete patched document. See the [pag
...
+
+```
+{
+  "name": "invalid",
+  "data": {},
+  "message": "You must delete the children of this page first."
```
localghost443, line 414: No child separation

docs/reference/api/pieces.md
```diff
@@ -230,11 +230,43 @@ The PATCH request body may use MongoDB-style operators. For example, you may use
 
 The successful PATCH request returns the complete patched document. See the [piece document response example](#piece-document-response-example) below for a sample response body. On error an appropriate HTTP status code is returned.
```
localghost443, line 231: Is this a link to something immediately beneath this text? "piece document response example"

docs/reference/api/pieces.md
```diff
@@ -230,11 +230,43 @@ The PATCH request body may use MongoDB-style operators. For example, you may use
...
 
+## `DELETE /api/v1/:piece-name/:_id`
+
+**Authentication required.**
+
+This API route **permanently deletes the piece database document**. Moving pieces to the trash in the Apostrophe user interface or using a PATCH request to set `trash: true` do not permanently delete database documents and are typically better options.
```
localghost443, line 237: Same with better options here

docs/reference/api/pieces.md
```diff
@@ -230,11 +230,43 @@ The PATCH request body may use MongoDB-style operators. For example, you may use
...
+```javascript
+// Request inside an async function.
+await fetch('http://example.net/api/v1/article/ckitdo5oq004pu69kr6oxo6fr:en:published?apikey=myapikey', {
+  method: 'DELETE'
+});
+```
+
+### Response
+
+The successful DELETE request simply responds with a `200` HTTP response status code. On error an appropriate HTTP status code is returned.
```
localghost443, line 261: Same with "on error" here

localghost443: and DELETE

docs/reference/api/pages.md
abea: On the first point, this sentence is part of a normal paragraph (unlike the previous line), so I was trying to maintain normal grammar, hence the subject.

docs/reference/api/pages.md
abea: I think that's still proper syntax. It's an alternate name for the thing and I didn't want to only use "children" for folks who might not be as familiar with the usage.

docs/reference/api/pages.md
abea: We write REST request names like that elsewhere, but as I'm looking at our peers it does looks like the types should be written as code. I'll edit those usages.

docs/reference/api/pieces.md
abea: Yeah, that's further down the page.

docs/reference/api/authentication.md
```diff
@@ -71,11 +71,11 @@ A successful response will return a JSON object with a `token` property.
 }
 ```
 
-On error an appropriate HTTP status code is returned.
+In case of an error an appropriate HTTP status code is returned.
 
 ### Usage
 
-Add this token to all REST API requests that require login by passing an `Authorization` header. This includes POST, PUT, PATCH and DELETE operations. You will also get more complete results from GET operations, including access to documents with `visibility` set to `loginRequired`.
+Add this token to all REST API requests that require login by passing an `Authorization` header. This includes `POST`, `PUT`, `PATCH` and `DELETE` operations. You will also get more complete results from `GET` operations, including access to documents with `visibility` set to `loginRequired`.
```
localghost443, line 78: receive vs get

docs/reference/api/authentication.md
```diff
@@ -85,12 +85,12 @@ Bearer random123Token456xyz
 
 Note the need for the word "Bearer" before the key.
 
-To log out and destroy the token, send a POST request to `/api/v1/login/logout`, with the same `authorization` header. No body properties are required. After logging out, the token is no longer accepted.
+To log out and destroy the token, send a `POST` request to `/api/v1/login/logout`, with the same `authorization` header. No body properties are required. After logging out, the token is no longer accepted.
```
localghost443, line 88: Will no longer be accepted

docs/reference/api/pages.md
```diff
@@ -314,13 +314,13 @@ const document = await response.json();
...
 
-The successful PUT request returns the newly created document. See the [page document response example](#page-document-response-example) below for a sample response body. On error an appropriate HTTP status code is returned.
+The successful `PUT` request returns the newly created document. See the [page document response example](#page-document-response-example) below for a sample response body. In case of an error an appropriate HTTP status code is returned.
 
 ## `PATCH /api/v1/@apostrophecms/page/:_id`
 
 **Authentication required.**
 
-The PATCH request may include *both* `_targetId` and `_position` as described in the [POST request description](#post-api-v1-apostrophecms-page), but that is not required if the page is not being moved.
+The `PATCH` request may include *both* `_targetId` and `_position` as described in the [`POST` request description](#post-api-v1-apostrophecms-page), but that is not required if the page is not being moved.
```
localghost443, line 323: This is a double negative “not required if x is not being moved”

docs/reference/api/authentication.md
localghost443: Maybe not actually

localghost443 (approved):

PR #38, opened 2021-02-25
Module options

abea:
Documents module options common to _all_ modules (with limited exceptions).

Especially looking for feedback on:
- introducing this feature area
- technical accuracy
- format


docs/reference/module-options.md
```diff
@@ -0,0 +1,148 @@
+# Module configuration options
+
+Most Apostrophe modules, both core and project-level, can be configured with settings that influence functionality without having to write custom Javascript. The sections below describe the options available in all modules as well as those specific to certain module types.
```
localghost443, line 3: I would strike "both core and project-level". If the truth is "most", is it worthwhile to carve out the exceptions?

docs/reference/module-options.md
```diff
@@ -0,0 +1,148 @@
...
+
+All the settings described here are placed in a module's `options` configuration object. This often added in the respective `index.js` file, but can also be added where the module is instantiated in the `app.js` file.
```
localghost443, line 5: "The `options` object can be added in the module's `index.js` file, as well as where the module is instantiated in the `app.js` file.

Remove "the" in "All the settings"

docs/reference/module-options.md
```diff
@@ -0,0 +1,148 @@
...
+
+`index.js` example:
+```javascript
+// In `modules/article/index.js`
```
localghost443, line 9: I think we're sometimes inconsistent here. Sometimes we just put the location at the very top of the code block like

`// modules/article/index.js`

Here, we have "In" with quotes.

I prefer the former, but thing consistency is the most important

docs/reference/module-options.md
```diff
@@ -0,0 +1,148 @@
...
+
+- [`alias`](#alias)
+- [`browser`](#browser)
+<!-- - [`components`](#components) -->
+- [`name`](#name)
+- [`templateData`](#templatedata)
+
+### `alias`
+
+Set to a string, the `alias` value will be applied to the `apos` object (accessible in many places) as a quick reference to the module.
```
localghost443, line 54: Should we be linking to `apos` object? Can we?

docs/reference/module-options.md
```diff
@@ -0,0 +1,148 @@
...
+
+<!-- ### `components`
+
+  The `components` option is used to identify Vue components
+
+  ~ set specific Vue components to be used for this module. (Don't doc?) -->
+
+### `name`
+
+Modules `name` properties are based on the `module` object keys in `app.js` by default. Setting the the `name` option in a module can override this original value for the sake of referencing the module elsewhere.
```
localghost443, line 104: Module vs Modules?

docs/reference/module-options.md
```diff
@@ -0,0 +1,148 @@
...
+  // ...
+}
+```
+
+You might use that value as a fallback for user-editable fields.
+
+```django
+{# In `modules/heading-widget/index.js` #}
+{% set bgColor = data.widget.color or data.defaultColor %}
+<h2 style="background-color: {{ bgColor }}">
```
localghost443, line 143: I think this is fine, but it's a strange example (using inline CSS). Should we do something that would be more recommended practice?

docs/reference/module-options.md
abea: We don't have that yet...

docs/reference/module-options.md
abea: I left the apostrophe off. Ironic.

docs/reference/module-options.md
abea: I don't know what a great one would be. @boutell ? As I was thinking about it, it did seem like this could be a better option for default values than hard-coding them into templates like we sometimes do. This way the default values are in the same file (index.js, probably) as the field options.

boutell: I don't think inline CSS is a bad practice for fields of type color.

docs/reference/module-options.md
boutell: "With the exception of modules present in the apostrophe core at the time of the 3.x stable release, modules distributed via npm should never set alias for themselves or assume that it has been set."

docs/reference/module-options.md
```diff
@@ -0,0 +1,139 @@
...
+  // ...
+}
+```
+
+This `article` module can then be referenced anywhere the `apos` object is present as `apos.article`. Otherwise it would be available via `apos.modules['news-article']`.
+
+
+### `components`
+
+The `components` options is an object identifying Vue components to use for the module's related user interface. The keys of this object, and thus the UI being overridden, will vary based on the module type. For example, piece module use `managerModal` and `editorModal` components.
```
boutell, line 75: piece modules

docs/reference/module-options.md
```diff
@@ -0,0 +1,139 @@
...
+      managerModal: 'MyCustomPiecesManager'
+    }
+  },
+  // ...
+}
+```
+
+### `name`
+
+Modules' `name` properties are based on the `module` object keys in `app.js` by default. Setting the the `name` option in a module can override this original value for the sake of referencing the module elsewhere.
```
boutell, line 95: The purpose of `name` has historically been to set the string used as a `type` property for docs of that type. In 3.x it is by default always the same as the module name, and I think we should stomp out `self.options.name` in favor of setting `self.name` firmly to `self.__meta.name` rather than just as a default. It can only cause confusion.

boutell (changes requested):

docs/reference/module-options.md
abea: Added this in a note below.

docs/reference/module-options.md
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+});
+```
+
+## Options for any module
+
+- [`alias`](#alias)
+- [`components`](#components)
+- [`name`](#name)
```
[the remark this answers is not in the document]
abea: Should we remove the ability to set the name option as a default override, then? I don't know that I understand what you're saying.

docs/reference/module-options.md
abea: that works for me.

docs/reference/module-options.md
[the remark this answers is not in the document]
abea: Ah, you clarified in your other comment. That works. I'll remove from here and plan on excising it from core.

boutell (approved):

docs/reference/module-options.md
localghost443: I suggest using a different field type, a more common one.

PR #39, opened 2021-03-01
Module options 2

abea:
Adds additional module option documentation. Specifically covers doc type and piece type options, with examples. Also other related documentation improvements.

### DIFF
--- a/docs/reference/api/pages.md
+++ b/docs/reference/api/pages.md
@@ -322,8 +322,6 @@ The successful `PUT` request returns the newly created document. See the [page d
 
 ## `PATCH /api/v1/@apostrophecms/page/:_id`
 
-The `PATCH`request may include *both* `_targetId` and `_position` as described in the [POST request description](#post-api-v1-apostrophecms-page). That is only required when moving the page.
-
 ### Query parameters
 
 | Parameter | Example | Description |
@@ -331,6 +329,8 @@ The `PATCH`request may include *both* `_targetId` and `_position` as described i
 |`apos-mode` | `?apos-mode=draft` | Set to `draft` or `published` to update a specific mode version of the page. |
 |`apos-locale` | `?apos-locale=fr` | Set to a valid locale to update the page document version for that locale. |
 <!-- TODO: link to docs about locales when available. -->
+
+If moving a page within the page tree, the `PATCH`request must include *both* `_targetId` and `_position` as described in the [POST request description](#post-api-v1-apostrophecms-page).
 
 If a `PATCH` operation is attempted in the published mode, the changes in the patch are applied to both the draft and the current document, but properties of the draft not mentioned in the patch are not published. This is to prevent unexpected outcomes.
 
--- a/docs/reference/api/pages.md
+++ b/docs/reference/api/pages.md
@@ -1,6 +1,6 @@
 # Page type REST endpoints
 
-All [page types](/reference/glossary.md#page) use a single set of API endpoints, unlike piece types. Difference in page type will primarily influence returned document properties based on the page type configurations.
+All [page types](/reference/glossary.md#page) use a single set of API endpoints, unlike piece types.
 
 ## Endpoints
 
@@ -10,8 +10,8 @@ All [page types](/reference/glossary.md#page) use a single set of API endpoints,
 
 | Method | Path | Description |
 |---------|---------|---------|
-|`GET` | [`/api/v1/@apostrophecms/page` ](#get-api-v1-apostrophecms-page)| Get the home page and all other pages structured in the home page's `_children` property |
-|`GET` | [`/api/v1/@apostrophecms/page/:_id`](#get-api-v1-apostrophecms-page-id) | Get a single page with a specified ID |
+|`GET` | [`/api/v1/@apostrophecms/page` ](#get-api-v1-apostrophecms-page)| Fetch the home page and all other pages structured in the home page's `_children` property |
+|`GET` | [`/api/v1/@apostrophecms/page/:_id`](#get-api-v1-apostrophecms-page-id) | Fetch a single page with a specified ID |
 |`POST` | [`/api/v1/@apostrophecms/page`](#post-api-v1-apostrophecms-page) | Insert a new page |
 |`PUT` | [`/api/v1/@apostrophecms/page/:_id`](#put-api-v1-apostrophecms-page-id) | Fully replace a specific page database document |
 |`PATCH` | [`/api/v1/@apostrophecms/page/:_id`](#patch-api-v1-apostrophecms-page-id) | Update only certain fields on a specific page |
@@ -322,7 +322,7 @@ The successful `PUT` request returns the newly created document. See the [page d
 
 ## `PATCH /api/v1/@apostrophecms/page/:_id`
 
-The `PATCH`request may include *both* `_targetId` and `_position` as described in the [POST request description](#post-api-v1-apostrophecms-page), but that is not required if the page is not being moved.
+The `PATCH`request may include *both* `_targetId` and `_position` as described in the [POST request description](#post-api-v1-apostrophecms-page). That is only required when moving the page.
 
 ### Query parameters
 
--- a/docs/reference/api/pages.md
+++ b/docs/reference/api/pages.md
@@ -1,33 +1,35 @@
 # Page type REST endpoints
 
-All [page types](#TODO) use a single set of API endpoints, unlike piece types. Difference in page type will primarily influence returned document properties based on the page type configurations.
+All [page types](/reference/glossary.md#page) use a single set of API endpoints, unlike piece types. Difference in page type will primarily influence returned document properties based on the page type configurations.
 
 ## Endpoints
 
+[Authentication](/reference/api/authentication.md) is required for all REST API requests.
+
 ### REST endpoints
 
-| Method | Path | Description | Auth required |
-|---------|---------|---------|---------|
-|GET | [`/api/v1/@apostrophecms/page` ](#get-api-v1-apostrophecms-page)| Get the home page and all other pages structured in the home page's `_children` property | FALSE |
-|GET | [`/api/v1/@apostrophecms/page/:_id`](#get-api-v1-apostrophecms-page-id) | Get a single page with a specified ID | FALSE |
-|POST | [`/api/v1/@apostrophecms/page`](#post-api-v1-apostrophecms-page) | Insert a new page | TRUE |
-|PUT | [`/api/v1/@apostrophecms/page/:_id`](#put-api-v1-apostrophecms-page-id) | Fully replace a specific page database document | TRUE |
-|PATCH | [`/api/v1/@apostrophecms/page/:_id`](#patch-api-v1-apostrophecms-page-id) | Update only certain fields on a specific page | TRUE |
-|DELETE | Not supported | [See more in PATCH request details.](#moving-pages-to-the-trash) | n/a |
+| Method | Path | Description |
+|---------|---------|---------|
+|GET | [`/api/v1/@apostrophecms/page` ](#get-api-v1-apostrophecms-page)| Get the home page and all other pages structured in the home page's `_children` property |
+|GET | [`/api/v1/@apostrophecms/page/:_id`](#get-api-v1-apostrophecms-page-id) | Get a single page with a specified ID |
+|POST | [`/api/v1/@apostrophecms/page`](#post-api-v1-apostrophecms-page) | Insert a new page |
+|PUT | [`/api/v1/@apostrophecms/page/:_id`](#put-api-v1-apostrophecms-page-id) | Fully replace a specific page database document |
+|PATCH | [`/api/v1/@apostrophecms/page/:_id`](#patch-api-v1-apostrophecms-page-id) | Update only certain fields on a specific page |
+|DELETE | Not supported | [See more in PATCH request details.](#moving-pages-to-the-trash) |
 
 ### Additional page endpoints
 
-| Method | Path | Description | Auth required |
+| Method | Path | Description |
 |---------|---------|---------|---------|
-|GET | [`/:_url?apos-refresh=1`](#get-url-apos-refresh-1) | Get a page's rendered content | FALSE |
-|POST | [`/api/v1/@apostrophecms/page/:_id/publish`](#post-api-v1-apostrophecms-page-id-publish) | Publish the draft version of a page | TRUE |
+|GET | [`/:_url?apos-refresh=1`](#get-url-apos-refresh-1) | Get a page's rendered content |
+|POST | [`/api/v1/@apostrophecms/page/:_id/publish`](#post-api-v1-apostrophecms-page-id-publish) | Publish the draft version of a page |
 
 <!-- TODOcument -->
-<!-- |GET | [`/api/v1/@apostrophecms/page/:_id?locales=1`](#get-api-v1-apostrophecms-page-id-locales-1) | Request the available locales for a specific page | TRUE |
-|POST | [`/api/v1/@apostrophecms/page/:_id/export`](#post-api-v1-apostrophecms-page-id-export) | Copy a page from one locale to another | TRUE |
-|POST | [`/api/v1/@apostrophecms/page/:_id/revert-draft-to-published`](#post-api-v1-apostrophecms-page-id-revert-draft-to-published) | Revert a draft page to the the state of the published version, if available | TRUE |
-|POST | [`/api/v1/@apostrophecms/page/:_id/revert-published-to-previous`](#post-api-v1-apostrophecms-page-id-revert-published-to-previous) | Revert a published page to the previous version, if available | TRUE |
-|POST | [`/api/v1/@apostrophecms/page/:_id/unpublish`](#post-api-v1-apostrophecms-page-id-unpublish) | Unpublish the published version of a page | TRUE | -->
+<!-- |GET | [`/api/v1/@apostrophecms/page/:_id?locales=1`](#get-api-v1-apostrophecms-page-id-locales-1) | Request the available locales for a specific page |
+|POST | [`/api/v1/@apostrophecms/page/:_id/export`](#post-api-v1-apostrophecms-page-id-export) | Copy a page from one locale to another |
+|POST | [`/api/v1/@apostrophecms/page/:_id/revert-draft-to-published`](#post-api-v1-apostrophecms-page-id-revert-draft-to-published) | Revert a draft page to the the state of the published version, if available |
+|POST | [`/api/v1/@apostrophecms/page/:_id/revert-published-to-previous`](#post-api-v1-apostrophecms-page-id-revert-published-to-previous) | Revert a published page to the previous version, if available |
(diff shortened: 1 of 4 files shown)


docs/reference/api/pages.md
```diff
@@ -1,33 +1,37 @@
...
 
 ## Endpoints
 
+[Authentication](/reference/api/authentication.md) is required for all REST API requests.
+
 ### REST endpoints
 
-| Method | Path | Description | Auth required |
-|---------|---------|---------|---------|
-|`GET` | [`/api/v1/@apostrophecms/page` ](#get-api-v1-apostrophecms-page)| Get the home page and all other pages structured in the home page's `_children` property | FALSE |
```
abea, line 11: These big change blocks for the tables are from removing the "Auth required" column since it's required for all.

docs/reference/glossary.md
```diff
@@ -45,6 +45,15 @@ Show page templates (`show.html`), do have access to their parent index page's `
 
 There is a single doc with the slug, `global`, which is always loaded and available to page templates as `data.global`. This is useful for shared, site-wide headers and footers that are editable, etc. It is managed by the `@apostrophecms/global` module.
 
+## Slug
```
abea, line 48: 🆕 glossary entry

docs/reference/module-options.md
```diff
@@ -136,4 +138,257 @@ You might use that value as a fallback for user-editable fields.
...
+
+<!-- - [`createControls`](#createcontrols)
+- [`editControls`](#editcontrols) -->
+- [`pluralLabel`](#plurallabel)
+- [`perPage`](#perpage)
+- [`publicApiProjection`](#publicapiprojection)
+- [`quickCreate`](#quickcreate)
+- [`searchable`](#searchable)
+
+<!-- ### `createControls`
```
abea, line 293: Don't document these, right? They're pretty serious UI editing options.

boutell: createControls and editControls are dead code.

docs/reference/glossary.md
```diff
@@ -45,6 +45,15 @@ Show page templates (`show.html`), do have access to their parent index page's `
...
+
+The slug of an Apostrophe doc is a string that uniquely identifies that doc within the database. Apostrophe will enforce that no two docs may have the same slug. Unlike the `_id` property of the database document, slugs _can_ change, however. Apostrophe user interface and REST API will suggest slugs automatically based on the document's title on creation.
+
+If the doc has a dedicated page (or [show page](#show-page) for a piece), the slug is used as a component of the page URL. This works differently for pages and pieces.
```
boutell, line 52: If the doc is a page, or a piece with a show page...

docs/reference/glossary.md
```diff
@@ -45,6 +45,15 @@ Show page templates (`show.html`), do have access to their parent index page's `
...
+
+- [Page](#page) slugs begin with a forward slash and, by default, include sections for their parent pages, e.g., `/previous-page/new-page` for a page titled "New Page" with a parent page of "Previous Page." Editing a page slug will not effect the actual page tree structure, however.
+- [Piece](#piece) slugs do not begin with a forward slash nor do they include reference to an [index page](#index-page). A piece titled "New Piece," will get the suggested slug of `new-piece` if no other doc has that slug already.
```
boutell, line 55: Lose the comma.

docs/reference/module-options.md
```diff
@@ -136,4 +138,257 @@ You might use that value as a fallback for user-editable fields.
 </h2>
 ```
 
-## Accessing module options
\ No newline at end of file
+## Options for all doc type modules
+
+Option settings in this section apply to all modules that extend `@apostrophecms/doc-type` ([doc type](glossary.md#doc) modules). These include all piece and page types.
+
+- [`adminOnly`](#adminonly)
```
boutell, line 145: Only users with the sitewide admin permission can read or write docs of this type. `@apostrophecms/user` itself is a good example.

docs/reference/module-options.md
```diff
@@ -136,4 +138,257 @@ You might use that value as a fallback for user-editable fields.
...
+  options: {
+    adminOnly: true
+  },
+  // ...
+}
+```
+
+### `autopublish`
+
+Set `autopublish` to `true` to automatically publish any changes saved to docs of this type. There is then effectively no draft mode for this doc type. The core image and file modules use this option, for example, and it can be useful for such "utility" piece types that need to have a single, predictable state.
```
boutell, line 173: This option eliminates the need for users to think about the distinction between draft and published content while preserving the possibility of translation for different locales. That is not quite a "single, predictable state." It is only a user interface convenience.

docs/reference/module-options.md
```diff
@@ -136,4 +138,257 @@ You might use that value as a fallback for user-editable fields.
...
+  options: {
+    label: 'Featured Article'
+  },
+  // ...
+}
+```
+
+### `localized`
+
+Defaults to `true`. If set to `false`, this doc type will _not_ be included in the locale system. This means there will be only one version of each doc, regardless of whether multiple locales (e.g., for languages or regions) are active. The "users" piece disables localization in this way.
```
boutell, line 215: `@apostrophecms/user`

This is the "single predictable state" option.

boutell (changes requested):

docs/reference/glossary.md
abea: better.

docs/reference/module-options.md
```diff
@@ -136,4 +138,257 @@ You might use that value as a fallback for user-editable fields.
...
+
+#### Example
+
+```javascript
+// modules/goose/index.js
+module.exports = {
+  extend: '@apostrophecms/piece-type',
+  options: {
+    label: 'Goose',
+    pluralLabel: 'Geese'
```
abea, line 312: I was upset there's no goose emoji.

docs/reference/module-options.md
abea: description updated below

boutell (approved):

docs/reference/api/pages.md
```diff
@@ -1,33 +1,37 @@
 # Page type REST endpoints
 
-All [page types](#TODO) use a single set of API endpoints, unlike piece types. Difference in page type will primarily influence returned document properties based on the page type configurations.
+All [page types](/reference/glossary.md#page) use a single set of API endpoints, unlike piece types. Difference in page type will primarily influence returned document properties based on the page type configurations.
```
localghost443, line 3: Might just be me, but the second sentence seems....needlessly complicated? Feels like we're saying something very simple with a lot of words.

docs/reference/api/pages.md
```diff
@@ -1,33 +1,37 @@
...
-|---------|---------|---------|---------|
-|`GET` | [`/api/v1/@apostrophecms/page` ](#get-api-v1-apostrophecms-page)| Get the home page and all other pages structured in the home page's `_children` property | FALSE |
-|`GET` | [`/api/v1/@apostrophecms/page/:_id`](#get-api-v1-apostrophecms-page-id) | Get a single page with a specified ID | FALSE |
-|`POST` | [`/api/v1/@apostrophecms/page`](#post-api-v1-apostrophecms-page) | Insert a new page | TRUE |
-|`PUT` | [`/api/v1/@apostrophecms/page/:_id`](#put-api-v1-apostrophecms-page-id) | Fully replace a specific page database document | TRUE |
-|`PATCH` | [`/api/v1/@apostrophecms/page/:_id`](#patch-api-v1-apostrophecms-page-id) | Update only certain fields on a specific page | TRUE |
-|`DELETE` | [`/api/v1/@apostrophecms/page/:_id`](#delete-api-v1-apostrophecms-page-id) | **Permanently deletes a page document** | n/a |
+| Method | Path | Description |
+|---------|---------|---------|
+|`GET` | [`/api/v1/@apostrophecms/page` ](#get-api-v1-apostrophecms-page)| Get the home page and all other pages structured in the home page's `_children` property |
```
localghost443, line 13: We're using 'get' almost everywhere so I think its fine, but I've long been wondering if 'fetch' is better. "Get"....is literately `GET`, but it's also not quite...correct. Idk. Feel free to ignore.

docs/reference/api/pages.md
```diff
@@ -241,7 +245,8 @@ The `_position` property uses specific string values rather than index numbers t
 | Parameter | Example | Description |
 |----------|------|-------------|
 |`apos-mode` | `?apos-mode=draft` | Set to `draft` to insert a page as a draft instead of immediately published. Set to `published` or leave it off to insert a published page. |
-|`apos-locale` | `?apos-locale=fr` | Set to [a valid locale](#TODO) to request page document versions for that locale. Defaults to the default locale. |
+|`apos-locale` | `?apos-locale=fr` | Set to a valid locale to request page document versions for that locale. Defaults to the default locale. |
+<!-- TODO: link to docs about locales when available. -->
```
localghost443, line 249: :)

docs/reference/api/pages.md
```diff
@@ -318,16 +322,15 @@ The successful `PUT` request returns the newly created document. See the [page d
 
 ## `PATCH /api/v1/@apostrophecms/page/:_id`
 
-**Authentication required.**
-
-The `PATCH` request may include *both* `_targetId` and `_position` as described in the [`POST` request description](#post-api-v1-apostrophecms-page), but that is only required if the page is being moved.
+The `PATCH`request may include *both* `_targetId` and `_position` as described in the [POST request description](#post-api-v1-apostrophecms-page), but that is not required if the page is not being moved.
```
localghost443, line 325: Is there a way to avoid the double negative here?

docs/reference/api/pieces.md
```diff
@@ -1,27 +1,29 @@
 # Piece type REST endpoints
 
-Every [piece type](#TODO) has built in REST end points that share their overall structure in common. The exact document properties returned will depend on the piece type's fields.
+Every [piece type](/reference/glossary.md#piece) has built in REST end points that share their overall structure in common. The exact document properties returned will depend on the piece type's fields.
```
localghost443, line 3: "has built in REST end points that share their overall structure in common". I suggest rephrasing. Its not clear what you're trying to say here.

docs/reference/glossary.md
```diff
@@ -45,6 +45,15 @@ Show page templates (`show.html`), do have access to their parent index page's `
 
 There is a single doc with the slug, `global`, which is always loaded and available to page templates as `data.global`. This is useful for shared, site-wide headers and footers that are editable, etc. It is managed by the `@apostrophecms/global` module.
 
+## Slug
+
+The slug of an Apostrophe doc is a string that uniquely identifies that doc within the database. Apostrophe will enforce that no two docs may have the same slug. Unlike the `_id` property of the database document, slugs _can_ change, however. Apostrophe user interface and REST API will suggest slugs automatically based on the document's title on creation.
```
localghost443, line 50: "no two docs may" remove "may"

"slugs _can_ change, however", remove "however". Also consider reversing. "Slugs can change, unlike the `id`", for example. (https://www.learngrammar.net/english-grammar/comparatives)

localghost443: **The** Apostrophe user interface

docs/reference/glossary.md
```diff
@@ -45,6 +45,15 @@ Show page templates (`show.html`), do have access to their parent index page's `
...
+
+If the doc is a page, or a piece with a [show page](#show-page), the slug is used as a component of the page URL. This works differently for pages and pieces.
+
+- [Page](#page) slugs begin with a forward slash and, by default, include sections for their parent pages, e.g., `/previous-page/new-page` for a page titled "New Page" with a parent page of "Previous Page." Editing a page slug will not effect the actual page tree structure, however.
```
localghost443, line 54: However not necessary

docs/reference/module-options.md
```diff
@@ -136,4 +138,252 @@ You might use that value as a fallback for user-editable fields.
...
+  options: {
+    localized: false
+  },
+  // ...
+}
+```
+
+### `sort`
+
+The `sort` option for a doc type defines a sorting order for requests to the database for that type. This setting generally follows the [MongoDB `$sort` aggregation](https://docs.mongodb.com/manual/reference/operator/aggregation/sort/) syntax. The option is set to an object containing field name keys with `1` as a property value for ascending order and `-1` for descending order.
```
localghost443, line 234: While its useful to reference the mongoldb doc here, I think it would be more worthwhile just to show these options in our own table, rather than linking to the mongo table and then listing them out.

docs/reference/module-options.md
```diff
@@ -136,4 +138,252 @@ You might use that value as a fallback for user-editable fields.
...
+      // ...
+    }
+  }
+  // ...
+}
+```
+
+### `slugPrefix`
+
+Set `slugPrefix` to a string to prepend all [slugs](glossary.md#slug) for docs of this type. This can be useful to help prevent slugs, which must be unique for each doc in the database, from being reserved in some cases. For example, Apostrophe image docs have the `slugPrefix` value of `'image-'` so images, which do not typically have public pages, do not accidentally reserve a more reader-friendly slug.
```
localghost443, line 268: "This can be useful to help prevent slugs, which must be unique for each doc in the database, from being reserved in some cases" to "This can prevent slugs from being reserved in some cases".

docs/reference/module-options.md
```diff
@@ -136,4 +138,252 @@ You might use that value as a fallback for user-editable fields.
...
+
+- [`pluralLabel`](#plurallabel)
+- [`perPage`](#perpage)
+- [`publicApiProjection`](#publicapiprojection)
+- [`quickCreate`](#quickcreate)
+- [`searchable`](#searchable)
+
+### `pluralLabel`
+
+Similar to `label` for all doc types, the `pluraLabel` option sets the string the user interface will use to describe a piece type in plural contexts. All page types are referred to as "Pages" in these contexts, but pieces have unique labels, for example, in the manager modal where it might display "All Articles."
```
localghost443, line 295: ", for example, in the manager modal where it might display "All Articles."

Suggesting we end the previous sentence and begin a new sentence at "For example". Also suggest we remove references to manager modal and simply say "For example, article becomes 'articles'." or something to that effect.

docs/reference/module-options.md
localghost443: great example tho.

docs/reference/module-options.md
```diff
@@ -136,4 +138,252 @@ You might use that value as a fallback for user-editable fields.
...
+    label: 'Goose',
+    pluralLabel: 'Geese'
+  },
+  // ...
+}
+```
+
+### `perPage`
+
+In piece type modules, the `perPage` option, set to an integer, sets the number of pieces that will be returned in each "page" [during `GET` requests](api/pieces.md#get-api-v1-piece-name) that don't specify an `_id`. This value defaults to 10.
```
localghost443, line 315: Still ruffling my feathers around the heavy handedness of "modules". In Pieces? For a piece type?

"set to an integer", "expressed as an integer"

docs/reference/module-options.md
```diff
@@ -136,4 +138,252 @@ You might use that value as a fallback for user-editable fields.
...
+  },
+  // ...
+}
+```
+
+Unauthenticated [`GET /api/v1/article`](api/pieces.md#get-api-v1-piece-name) requests would return each piece with only the title, `authorName`, and `_url` properties.
+
+### `quickCreate`
+
+Setting `quickCreate` to the boolean `true` on a piece adds that piece type to the admin bar "quick create" menu. The Apostrophe admin bar user interface includes the quick create menu button to add new pieces without first opening their respective manager modals. This is a useful setting for the piece types that editors will be adding most often.
```
localghost443, line 357: "to the boolean" `true`" Seems like a clunky way to say this. Wouldn't this be better expressed as an option|type|value table?

"This is a useful setting for the piece types that editors will be adding most often." Not necessary. We wax-on sometimes in ways I think makes it a bit more difficult to scan.

docs/reference/module-options.md
```diff
@@ -136,4 +138,252 @@ You might use that value as a fallback for user-editable fields.
...
+    quickCreate: true
+  },
+  // ...
+}
+```
+
+### `searchable`
+
+<!-- TODO: link to documentation of Apostrophe search when available. -->
+By default, all piece types are included in Apostrophe search results. Setting `searchable: false` on a piece type will exclude that piece type from the search results.
```
localghost443, line 375: First sentence is not necessary. I think again that this might be better expressed in a table?

docs/reference/api/pages.md
abea: i can remove it. it's not doing a lot.

docs/reference/api/pages.md
abea: I hear you. JS took all the words, "fetch" also being a technical JS term. "Return," also a technical term, might be good as well.

docs/reference/api/pieces.md
abea: I don't know either.

docs/reference/module-options.md
abea: I think it's worth making clear that they have to be unique, but I can trim from the starting clause.

docs/reference/module-options.md
abea: I can use "In piece types." In reference material I think it's less problematic/confusing to use "module" as it is in the guide. Still we can minimize it.

docs/reference/module-options.md
abea: A table might be useful for some of this. We could do it like the REST API pages with each section getting a leading table that links to the deeper descriptions. I'm thinking columns for option, value type (e.g., boolean), and brief description.

Maybe a good thing to do. This is where we could use some designer thinking. I'd also do that in a new PR if we're going to do it.

docs/reference/api/pages.md
```diff
@@ -1,33 +1,37 @@
...
+
 
 ### Additional page endpoints
 
-| Method | Path | Description | Auth required |
+| Method | Path | Description |
 |---------|---------|---------|---------|
-|`GET` | [`/:_url?apos-refresh=1`](#get-url-apos-refresh-1) | Get a page's rendered content | FALSE |
-|`POST` | [`/api/v1/@apostrophecms/page/:_id/publish`](#post-api-v1-apostrophecms-page-id-publish) | Publish the draft version of a page | TRUE |
+|`GET` | [`/:_url?apos-refresh=1`](#get-url-apos-refresh-1) | Get a page's rendered content |
```
localghost443, line 25: left a Get

docs/reference/api/pages.md
```diff
@@ -318,16 +322,15 @@ The successful `PUT` request returns the newly created document. See the [page d
 
 ## `PATCH /api/v1/@apostrophecms/page/:_id`
 
-**Authentication required.**
-
-The `PATCH` request may include *both* `_targetId` and `_position` as described in the [`POST` request description](#post-api-v1-apostrophecms-page), but that is only required if the page is being moved.
+The `PATCH`request may include *both* `_targetId` and `_position` as described in the [POST request description](#post-api-v1-apostrophecms-page). That is only required when moving the page.
```
localghost443, line 325: Instead of that, reference the phrase before. "Both are required only when"

docs/reference/api/pages.md
abea: There are a lot like this. If it's not a hard rule, let's try to avoid holding up PRs for things that are not from this PR.

localghost443 (approved):

PR #40, opened 2021-03-04
Module options 3

abea:
Adds options docs for page, page type, and piece page type modules.


docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
 }
 ```
 
+## Options for the core page module
+
+Option settings in this section apply to the core page module (`@apostrophecms/page`).
+
+| Option | Value type | Description |
```
abea, line 379: I'll go back and make them all tables later.

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+
+Option settings in this section apply to all piece page types (modules that extend `@apostrophecms/piece-page-type`).
+
+
+| Option | Value type | Description |
+|---------|---------|---------|
+| [`perPage`](#perpage) | Integer | Set the number of pieces to be |
+| [`next`](#next) | Boolean/Object | Enable and optionally configure the `req.data.next` object. |
+| [`piecesFilters`](#piecesfilters) | Array | Configure pieces filters for index pages. |
+| [`piecesModuleName`](#piecesmodulename) | String | Specify the associated piece type. |
```
boutell, line 650: Document the default behavior, this option is rarely needed.

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+  },
+  // ...
+}
+```
+
+### `next`
+
+If set to `true`, Apostrophe will include the next piece, based on the [sort option](#sort), on `req.data.next` when serving a [show page](/reference/glossary.md#show-page). This is useful to add links to a show page directing visitors to the next item in a series (e.g., the next oldest blog post). If not set, `req.data.next` will not be available.
+
+`next` can also be set to an object, which will be used as a query builder for retrieving that next piece document.
```
boutell, line 677: ... Yeah. Let's let this language stand. Sure, a "query builder" can refer to such an object containing query builder method names and values to pass to them. We can use the phrase "query builder method" to refer to the individual methods when there is ambiguity.

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+        // Category choices here
+      ]
+    }
+    // Other fields...
+  },
+  // ...
+}
+```
+
+### `piecesModuleName`
```
boutell, line 755: This option should get renamed `pieceModuleName`.

boutell: For consistency with our singular-everywhere policy on module names.

boutell (changes requested):

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
 
+## Options for the core page module
+
+Option settings in this section apply to the core page module (`@apostrophecms/page`).
+
+| Option | Value type | Description |
+|---------|---------|---------|
+| [`builders`](#builders) | Object | Set query builder values to be used when pages are served. |
+| [`home`](#home) | Boolean/Object | Change how the home page is added to `req.data` when pages are served. |
+| [`minimumPark`](#minimumpark) | Array | Override home page default properties when created. |
```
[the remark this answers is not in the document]
abea: This description is meant to be short and incomplete. Since the trash page is registered as a child of the home page I phrased it like this. The trash page is included in the full documentation below.

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+| [`park`](#park) | Array | Set pages to be created on site start with configuration. |
```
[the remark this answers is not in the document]
abea: "with additional configuration"? I don't know about that for this short description. Additional to what?

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
 }
 ```
 
+## Options for the core page module
```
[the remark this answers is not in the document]
abea: I was just thinking about this. We have the sidebar to make for easier navigation. Some of these modules only have one or two options, so a full page seems excessive. Ultimately we'll be doing a design process for docs to work some of this out.

docs/reference/module-options.md
abea: ➕

.styleguide.md
```diff
@@ -22,3 +22,38 @@ Javascript file example:
 
 Good: Make a `GET` request.
```
localghost443, line 23: What is this?

docs/reference/module-options.md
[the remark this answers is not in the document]
localghost443: " Set pages to be created on site **start** with configuration." Would this be site **creation**?

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+| [`park`](#park) | Array | Set pages to be created on site start with configuration. |
+| [`publicApiProjection`](#publicapiprojection-for-pages) | Object | Set query builder values to be used when pages are served. |
+| [`quickCreate`](#quickcreate-for-pages) | Boolean | Set to `false` to remove pages from the quick create menu. |
+| [`types`](#types) | Array | Set the page types available for new pages. |
+
+### `builders`
+
+<!-- TODO: Update link to a more detailed explanation of builders when
+available. -->
+The `builders` option can be used to apply any existing [query builders](/guide/major-changes.md#queries) when a page is served by its URL. This effects the data available on `req.data.page` (`data.page` in templates). This can be used to get more or less included on that page object depending on the need.
```
localghost443, line 393: "This can be used to get more or less included on that page object depending on the need.". Can we rephrase "this can be used to get more or less....". "This can be used to control what is included in the page object"

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+
+The default value is:
+```javascript
+{
+  children: true,
+  ancestors: { children: true }
+}
+```
+
+This includes one level of the page's page tree "children" as `_children` and its "ancestor" pages each with one level of their child pages on `_ancestors`.
```
localghost443, line 403: "This includes" -> "In this example," like you have below seems nice

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+}
+```
+
+In this example, we are not including ancestor pages and are requesting two levels of child pages (direct children and their direct children).
+
+### `home`
+
+The home page document is added to all page requests on `req.data.home` so it can be referenced in all page templates. That home page object also includes a `_children` property containing an array of top level page objects. There are two settings for this option to change that behavior. These can offer minor performance improvements for large sites.
+
+- Set `home: false` to disable adding the home page document to the requests.
```
localghost443, line 426: Can we do as table? If not ignore. If so, we can remove the statement on "two settings" and note the performance improvements in the table.

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+  // ...
+}
+```
+
+### `park`
+
+Use the `park` option to add an array of pages that should be created when the app starts up if they do not already exist. Each page is added as an object with initial properties, including the required `parkedId`.
+
+Required and recommended parked page properties include:
+- `parkedId` (required): A unique ID value used to identify it among parked pages.
```
localghost443, line 509: Table?

docs/reference/module-options.md
```diff
@@ -372,3 +372,442 @@ module.exports = {
...
+}
+```
+
+### `piecesFilters`
+
+<!-- TODO: Link to a guide on using piece filters when available. -->
+<!-- TODO: Link to a better query builder guide when available. -->
+`piecesFilters` can be configured as an array of objects to support filtering pieces on an [index page](/reference/glossary.md#index-page). Each object must have a `name` property associated with a valid [query builder](/guide/major-changes.md#queries). These include:
+
+- Field names whose field types automatically get builders: `string`, `slug`, `boolean`, `checkboxes`, `select`, `integer`, `float`, `url`, `date`, `relationship`
```
localghost443, line 716: Should at least be a list. I'm wondering if this is the...right place for this.

.styleguide.md
abea: It's formatting for REST type names.

docs/reference/module-options.md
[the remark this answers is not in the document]
abea: @localghost443 It gets semantic, but if a new page was added to this a month after launch, it would be created when the site was restarted, even though the site already was created.

docs/reference/module-options.md
abea: With the long description for the latter and only two options I thought bullets would be okay. It's sort of odd that there are two different but specific options like this, but there are reasons.

abea: Syke, it's fine.

docs/reference/module-options.md
abea: It should be in a guide about using filters for sure. I thought it'd be useful here as well.

docs/reference/module-options.md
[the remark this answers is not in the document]
boutell: I see, but it strongly implies it's only about the home page which isn't the case. How about:

"Override default parked pages, such as the home page."

docs/reference/module-options.md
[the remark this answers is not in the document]
boutell: Additional to the home page and trash page, but I think we could live without it.

docs/reference/module-options.md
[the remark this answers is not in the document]
boutell: OK. It's a good iterative approach. 👍

docs/reference/module-options.md
boutell: From Ben: "Set pages to be created on site start with configuration." Would this be site creation?"

Bea rightly points out that we can't give every detail in these quick summary tables and we provide more specifics later, but the specifics are:

Parked page settings are applied and, if necessary, re-applied when the process starts up. Also, Apostrophe's UI aims to prevent changes to any parked settings (TODO in A3). If you want to set defaults for certain properties but not enforce them again on every startup, use the `_defaults` subproperty of a parked page and specify defaults for individual fields as subproperties of `_defaults`.

docs/reference/module-options.md
boutell: I think it's worth it here because even between us the question of what piecesFilters is really good for kept coming up, due to a lack of visibility for this key feature: you always get builders "for free" for most schema fields.

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+
+The default value is:
+```javascript
+{
+  children: true,
+  ancestors: { children: true }
+}
+```
+
+In this example, page objects are loaded with one level of page tree "children" as `_children` and their "ancestor" pages, each with one level of their child pages, on `_ancestors`.
```
localghost443, line 403: Is there a better word for "loaded"

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+  },
+  // ...
+}
+```
+
+In this example, we are not including ancestor pages and are requesting two levels of child pages (direct children and their direct children).
+
+### `home`
+
+The home page document is added to all page requests on `req.data.home` so it can be referenced in all page templates. That home page object also includes a `_children` property containing an array of top level page objects. `home` option settings to offer minor performance improvements for large sites include:
```
localghost443, line 424: "Option settings for `home`, which offer minor performance improvements"

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+  options: {
+    home: { children: false }
+  },
+  // ...
+}
+```
+
+### `minimumPark`
+
+The `minimumPark` option offers the ability to overwrite how the default required pages are configured. This includes the home page and the trash "page" (trash is part of the page tree). Most "parking" pages should be done with the [`park`](#park) option.
```
localghost443, line 445: (trash is part of the page tree).....This is technically true right now, but there is a pre-launch ticket that seeks to separate these hierarchies, so I'm just flagging this as something we may need to circle back on before Alpha 7

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+
+::: warning
+Use with caution. Removing either of those pages as required with this option would have negative effects, but this does offer the ability to change the title or page type of the home page, for example.
```
localghost443, line 448: "Use with caution. Removing either of those pages as required with this option would have negative effects" This statement is a bit confusing. "Would have negative effects". We're also leading with the warning. It may make sense to lead with the function, and follow with the warning.

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+Required and recommended parked page properties include:
+
+| Setting | Requirement | Description |
+|---------|-------------|-------------|
+| `parkedId` | Required | A unique ID value used to identify it among parked pages. |
+| `slug` | Required | The page [slug](/reference/glossary.md#slug). |
+| `type` | Required | The page type to be used for the parked page. |
+| `title` | Recommended | The page title. If not set, it will be "New Page." |
+
+If added on the top level of the page object, these properties will not be editable through the user interface. Properties other than `parkedId` may be included in a `_defaults` property instead, which will allow them to be edited in the UI.
```
localghost443, line 519: admin interface

localghost443: maybe, maybe not

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+      _url: 1 // 👈 Dynamic properties are allowed
+    }
+  },
+  // ...
+}
+```
+
+Unauthenticated [`GET /api/v1/@apostrophecms/page`](api/pages.md#get-api-v1-apostrophecms-page) requests would return each piece with only the `title` and `_url` properties.
+
+### `quickCreate` (for pages)
```
localghost443, line 573: I'm assuming we cover quickCreate elsewhere too

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+  },
+  // ...
+}
+```
+
+### `piecesFilters`
+
+<!-- TODO: Link to a guide on using piece filters when available. -->
+<!-- TODO: Link to a better query builder guide when available. -->
+`piecesFilters` can be configured as an array of objects to support filtering pieces on an [index page](/reference/glossary.md#index-page). Each object must have a `name` property associated with a valid [query builder](/guide/major-changes.md#queries). These include:
```
localghost443, line 719: " can be configured as an array of objects" -> ",configured as an array of objects," unless there is some other way to do it

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+    }
+    // Other fields...
+  },
+  // ...
+}
+```
+
+### `pieceModuleName`
+
+Piece page types are each associated with a single piece type. If named with the pattern, `[piece name]-page`, the associated piece type will be detected automatically. For example, if the `article-page` module extends `@apostrophecms/piece-page-type`, it will automatically be associated with an `article` piece type.
```
localghost443, line 772: "If named with the pattern, `[piece name]-page`," maybe ditch the first comma?

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+
+This pattern can be overridden or ignored by explicitly setting `pieceModuleName` to an active piece type name. Ths can be useful if there is more than one piece page type for a single piece type (e.g., to support different functionality in each).
```
localghost443, line 774: "overridden" I think is past....."You can override or ignore this pattern by"

docs/reference/module-options.md
```diff
@@ -372,3 +372,457 @@ module.exports = {
...
+    pieceModuleName: 'person'
+  },
+  // Maybe there's code here to group people by team.
+  // ...
+}
+```
+
+### `previous`
+
+If set to `true`, Apostrophe will include the previous piece, based on the [sort option](#sort), on `req.data.previous` when serving a [show page](/reference/glossary.md#show-page). This is useful to add links to a show page directing visitors to the previous item in a series (e.g., the next newest blog post). If not set, `req.data.previous` will not be available.
```
localghost443, line 794: "This is useful to add links to a show page directing visitors to the previous item in a series (e.g., the next newest blog post). If not set, `req.data.previous` will not be available."

Feels like we should use "pagination" here (single page pagination) to simplify this sentence.

docs/reference/module-options.md
```diff
@@ -390,7 +390,7 @@ Option settings in this section apply to the core page module (`@apostrophecms/p
 
 <!-- TODO: Update link to a more detailed explanation of builders when
 available. -->
-The `builders` option can be used to apply any existing [query builders](/guide/major-changes.md#queries) when a page is served by its URL. This effects the data available on `req.data.page` (`data.page` in templates). This can be used to get more or less included on that page object depending on the need.
+The `builders` option can be used to apply any existing [query builders](/guide/major-changes.md#queries) when a page is served by its URL. This effects the data available on the page object, `req.data.page` (`data.page` in templates).
```
boutell, line 393: `affects`, not `effects`.

docs/reference/module-options.md
boutell: I think "fetched" is nice.

docs/reference/module-options.md
boutell: Should we change this to `pieceFilters` for consistency?

boutell (changes requested):

docs/reference/module-options.md
abea: Yeesh. I'll put a TODO on it rather than starting a deep dive on this now.

docs/reference/module-options.md
abea: Yeah, for pieces as well. But both are not included in "doc-type"

docs/reference/module-options.md
abea: How so? It's not part of our normal "pagination" functionality from index pages. I don't want to confuse it with that.

localghost443 (approved):

boutell (approved):

PR #41, opened 2021-03-04
Module options 4

abea:
The final group of module options and related clean up


docs/guide/module-format-example.md
```diff
@@ -377,6 +377,18 @@ module.exports = {
...
+  // Make SVG icons available for UI configuration if not already included
+  // in core. Icons must be available through the Material Design Icons (MDI)
+  // package: https://www.npmjs.com/package/@mdi/svg.
+  // These are configured as:
+  // 'reference-name-for-apostrophe': 'MaterialDesignIconsName'.
+  // The example below lets this and other modules use the MDI icon "AirHorn"
+  // using the keyword, `airhorn`.
+
+  icons: {
+    'airhorn': 'AirHorn'
```
localghost443, line 391: : )

docs/reference/module-options.md
```diff
@@ -41,13 +41,42 @@ require('apostrophe')({
 });
 ```
 
+## Using module options
+
+Most module options in core Apostrophe, and official modules, are used automatically for specific purposes. No additional work is needed to use them for their original purposes other than configuring them, if needed. In some cases it is helpful to reference the options object in custom module code. You can do this in any of the module configuration function sections using the `self` argument as `self.options`.
```
localghost443, line 46: "Most module options in core Apostrophe, and official modules"

All modules in this umbrella are "official", so I think the distinction is core/extended, not core/official. Not splitting hairs, just stumbled over this sentence a few times.

docs/reference/module-options.md
```diff
@@ -127,12 +156,14 @@ You might use that value as a fallback for user-editable fields.
...
-- [`label`](#label)
-- [`localized`](#localized)
-- [`sort`](#sort)
-- [`slugPrefix`](#slugprefix)
+| Option | Value type | Description |
+|---------|---------|---------|
+| [`adminOnly`](#adminonly) | Boolean | Set to `true` to only allow admins to manage the doc type. |
+| [`autopublish`](#autopublish) | Boolean | Set to `true` to publish all saved edits immediately. |
+| [`label`](#label-for-doc-types) | String | The readable label for the doc type. |
+| [`localized`](#localized) | Boolean | Set to `false` to not include the doc type in the locale system. |
```
localghost443, line 164: "not include" could be "exclude", or similar.

docs/reference/module-options.md
```diff
@@ -127,12 +156,14 @@ You might use that value as a fallback for user-editable fields.
...
-- [`autopublish`](#autopublish)
-- [`label`](#label)
-- [`localized`](#localized)
-- [`sort`](#sort)
-- [`slugPrefix`](#slugprefix)
+| Option | Value type | Description |
+|---------|---------|---------|
+| [`adminOnly`](#adminonly) | Boolean | Set to `true` to only allow admins to manage the doc type. |
+| [`autopublish`](#autopublish) | Boolean | Set to `true` to publish all saved edits immediately. |
+| [`label`](#label-for-doc-types) | String | The readable label for the doc type. |
```
localghost443, line 163: "readable" in what context? Is this human readable?

docs/reference/module-options.md
```diff
@@ -811,3 +843,198 @@ module.exports = {
...
 
+
+## Options for widget modules
+
+Option settings in this section apply to all widgets (modules that extend `@apostrophecms/widget-type`).
+
+
+| Option | Value type | Description |
+|---------|---------|---------|
+| [`className`](#classname) | String | A project-level class to apply to core widget templates. |
```
localghost443, line 854: I don't think we need to say 'project-level'. "Applies a class to core widget templates"

docs/reference/module-options.md
```diff
@@ -811,3 +843,198 @@ module.exports = {
...
+| [`icon`](#icon) | String | Select an available icon to include with the label in area menus. |
+| [`label`](#label-for-widgets) | String | The readable label for the widget type. |
+
+<!-- | [`scene`](#scene) | null | description | -->
+<!-- | [`contextual`](#contextual) | Boolean | description | -->
+
+### `className`
+
+Official Apostrophe widget templates support adding an html class from the `className` module option. The class is applied to the outer, wrapping HTML element in the widget template. This helps developers style the widget in project-level CSS using their namespacing and without targeting Apostrophe data attributes.
```
localghost443, line 863: "This helps developers".....We dont need this. we know html classes are for css. Right?

docs/reference/module-options.md
```diff
@@ -811,3 +843,198 @@ module.exports = {
...
+
+This option can be supported in custom, project-level widget templates (referenced with `data.manager.options.className` in template), but it may not be necessary. The option does not automatically have any effect in custom widget templates without use of `data.manager.options.className`.
```
localghost443, line 865: I'm confused by this statement. Help me out?

docs/reference/module-options.md
```diff
@@ -811,3 +843,198 @@ module.exports = {
...
+  // ...
+};
+
+```
+
+!['Area menu with icons next to widget labels'](/images/area-menu-with-icons.png)
+
+### `label` (for widgets)
+
+`label` should be set to a text string to be used in the area menu. If not set, Apostrophe will convert the module `name` meta property to a readable label by removing `-widget` from the end, splitting the `name` on dashes and underscores, then capitalizing the first letter of each word.
```
localghost443, line 932: "then captializing" to "and capitalizing"

docs/reference/module-options.md
```diff
@@ -811,3 +843,198 @@ module.exports = {
...
+
+| Option | Value type | Description |
+|---------|---------|---------|
+| [`defaultData`](#defaultdata) | Object | Define initial default data for rich text content. |
+| [`defaultOptions`](#defaultoptions) | Object | Configure the rich text toolbar and styles for rich text widgets. |
+| [`editorTools`](#editortools) | Object | Configure rich text tools and their Vue components. |
+
+### `defaultData`
+
+Rich text widgets can start with default content by setting `defaultData` to an object with a `content` property. That value would be a string of text or HTML that all rich text widgets would include when added.
```
localghost443, line 963: I didn't know we had this! Is this like, placeholder data? more or less?

docs/reference/module-options.md
```diff
@@ -811,3 +843,198 @@ module.exports = {
...
+      content: '<p>Replace me</p>'
+    }
+  },
+  // ...
+}
+```
+
+### `defaultOptions`
+
+The rich text widget is configured by default with useful [rich text toolbar settings and styles](https://github.com/apostrophecms/apostrophe/blob/3.0/modules/@apostrophecms/rich-text-widget/index.js#L15-L45). These can be overridden by setting project-level `defaultOptions`. This configuration object can include one or both of the `toolbar` and `styles` sub-options. If only one of those is included, the other will fall back to the core defaults.
```
localghost443, line 981: do we need project-level?

docs/reference/module-options.md
```diff
@@ -811,3 +843,198 @@ module.exports = {
...
+
+Project-level defaults can also be overridden in schema configuration where an area configures its rich text widgets. So a project can have site-wide defaults, but a specific area can have its own separate configuration.
```
localghost443, line 983: project-level. Isn't the assumption that we're in a project?

docs/reference/module-options.md
abea: well not all extended modules are official. But I can say "official extension modules" or something.

docs/reference/module-options.md
abea: it's confusing.

abea: i'll delete it. not helping much.

docs/reference/module-options.md
abea: Not exactly. Or sort of? If you leave it, it'll get saved and be public. So not placeholder in the sense that it's only visible to editors (like text input placeholders).

docs/reference/module-options.md
```diff
@@ -41,13 +41,42 @@ require('apostrophe')({
...
+module.exports = {
+  extend: '@apostrophecms/piece-type',
+  init (self) {
+    const moduleOptions = self.options;
+    // ...
+  },
+  methods (self) {
+    return {
+      logOptions () {
+        console.log('The module options are', self.options);
```
boutell, line 61: This will include the entire apos object, which arrives as an option too, so it's a pretty intense thing to log. Maybe log a specific example.

docs/reference/module-options.md
```diff
@@ -826,3 +858,196 @@ module.exports = {
...
+```
+
+<!-- TODO: Link to a guide page about configuring the RTE when available. -->
+
+### `editorTools`
+
+The rich text editor toolbar tools (e.g., "bold," "link," and "underline" buttons) can be reconfigured to have different labels (seen by assistive technologies), different icons, or even new Vue components altogether. `editorTools` can be completely overridden to do this if desired.
+
+::: warning
+Using this option takes full responsibility for the configuration of the rich text editor tools. Use this with caution. If overriding, be sure to include _all_ rich text tools that you will use.
```
boutell, line 1024: If your goal is to introduce an editor tool that does not yet exist, be aware that you will need to write a custom Vue component and often a tiptap extension as well. Refer to the Apostrophe module source for examples.

boutell (changes requested):

docs/reference/module-options.md
abea: I'm going to hold off on directing to the source code for something like this. I think it lets us off the hook to easily for now.

docs/reference/module-options.md
```diff
@@ -41,13 +41,42 @@ require('apostrophe')({
 });
 ```
 
+## Using module options
+
+Most module options in Apostrophe core and official extension modules are used automatically for specific purposes. No additional work is needed to use them for their original purposes other than configuring them, if needed. In some cases it is helpful to reference the options object in custom module code. You can do this in any of the module configuration function sections using the `self` argument as `self.options`.
```
localghost443, line 46: " No additional work is needed to use them for their original purposes other than configuring them, if needed." The 'if needed' feels weird here. No work is needed unless needed kind of vibe.

"In some cases it is helpful to reference the options object in custom module code." Are you saying its useful to reference our patterns when developing a custom module?

docs/reference/module-options.md
localghost443: Interesting. I love this though, we could definitely use it in Demos I think.

docs/reference/module-options.md
boutell: How about just:

"Your custom modules can access options of their own via `self.options`."

abea: @localghost443 Sometimes you want to reference the options object in custom module code. Your use of "patterns" is throwing me off.

I wish we had a better term for the "module configuration function sections."

abea: Updating to:

> Module options can also be referenced directly in custom module code. Module configuration function sections take a `self` argument, which is the module itself. You can then get the options as `self.options`.

localghost443 (approved):

boutell (approved):

PR #43, opened 2021-03-10
PRO-939 Module overview

abea:
Begins a page for module section reference documentation. This breaks the sections into two large groups: configuration settings and customization functions. This is new designation, but felt good to break the functions into their own section since they  have things in common.


docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+};
+```
+
+You should not use both `improve` and `extend` in a single module.
+
+### `improve`
+
+Similarly to `extend`, `improve` is used to name another existing module. Instead of _adopting_ the existing module's configuration, using `improve` _changes_ the named module's configuration.
+
+**This is only valid in modules that are built independently and then installed into Apostrophe apps**, either on their own or as a part of a bundle. It is most often used to improve core Apostrophe modules.
```
localghost443, line 63: Something not quite right with the phrasing of "this is only valid" and "built independently". What do both of those things mean?

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+
+### `fields`
+
+[Doc type](/reference/glossary.md#doc) modules have some fields configured by default, such as the `title` and `slug` fields. The `fields` setting is used for additional field management.
+
+The `fields` object is configured with subsections: `add`, `remove`, and `group`.
+
+#### `add`
+
+An object of fields to add to the doc type's schema. See the [field type reference](/reference/field-types/README.md) for more on field type configuration.
```
localghost443, line 91: I think simply 'schema' is fine.

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+        type: 'string'
+      }
+    }
+  }
+};
+```
+
+#### `remove`
+
+An array of field names from the base class module to remove. Some core doc type fields cannot be removed since they required by core functionality.
```
localghost443, line 109: We should say which fields cant be removed, or at least give a short list. Rather than saying 'required by for functionality',  I think we could say. "Some fields, such as [x] and [y] are required and can't be removed.

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+  extend: 'article',
+  fields: {
+    remove: [ 'subtitle' ]
+  }
+};
+```
+
+#### `group`
+
+An object of field groups. The groupings are used by editing interface. It _does not apply to widget modules_, which do not have a tabbed interface.
```
localghost443, line 123: "The" is not necessary. "Groupings are used by the editing interface"...."to organize sets of related fields"

"It does not apply" -> `group` does not apply.

Remove reference to 'tabbed interface'

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+
+Groups are added as an object with their name as the object key and the following properties:
+- `label`: The visible label (a string) on the interface tab
+- `fields`: An array of field names to include in the group
+
+The `@apostrophecms/doc-type` module groups the default fields in a `basics` group, which you can override. If overriding the `basics` group, you will need to add those core fields to a group in the module. Any fields not added to a group will be placed in an "Ungrouped" section in the editing interface.
```
localghost443, line 129: "group, which you can override. If overriding the `basics` group," 

remove ",which you can override". and try "You can override the `basics` group by adding the default fields to a group." 

We say 'the default' fields earlier and then call them core fields. I think default is better, but we should stick with it throughout the explanation. We may want to call out or link to 'default' fields in some way, if practical.

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+        ]
+      }
+    }
+  }
+};
+```
+
+### `filters`
+
+In piece type modules, the `filters` setting configures the pieces manager modal, adding and removing filtering interface (to view only certain pieces). `trash` and `visibility` filters are included by default.
```
localghost443, line 154: "the pieces manager modal" -- I think this is the first time in the docs ive seen this referenced. 

1. we might want to follow up with a definition or link at some point
2. Lets not call it a modal, its not relevant to describe its design, just its function.

",adding and removing filtering interface (to view only certain pieces)" -- this doesn't feel like a complete sentence.

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+
+The `filters` object is configured with subsections: `add` and `remove`. Filters must correspond to an existing fields name or custom [query builder](#queries-self) on the piece type.
+
+#### `add`
+
+An object of filters to add to the piece type. Each filter is an object with its own configuration. Filter properties include:
```
localghost443, line 160: Here and directly above, we simply say 'piece type' instead of 'piece type module' and this is good and I think we should do it more often. I asked in DMs, but I trip every time. Why isn't it 'piece-type'

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+
+| Property | Description |
+| ------- | ------- |
+| `label` | Recommended, but not required, to customize the human-readable label. |
+| `inputType` | Choose an input field type for the filter from the following: `radio`, `checkbox`, or `select`. Defaults to `select`. |
```
localghost443, line 165: "Choose an input field type for the filter from the following" -> "Declare the input field-type for the filter. `radio`, `checkbox`, or `select` are valid.  Defaults to `select`"

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+| `choices` | Manually set an array of choices instead of letting Apostrophe find the valid choices from the database. Choices include `label` and `value` properties. |
```
localghost443, line 166: "letting Apostrophe find the valid choices" -> "fetching choices"

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+  extend: 'article',
+  filters: {
+    remove: [ 'featured' ]
+  }
+};
+```
+
+### `columns`
+
+In piece type modules, the `columns` setting configures the pieces manager modal, adding and removing the piece listing columns. Default columns include `title`, `updatedAt`, and `visibility`.
```
localghost443, line 207: "In piece type modules" -> "For piece-types"?

"the piece listing columns" While its true that these are displayed in columns, its not columns they are removing, its the data/info that is displayed (in columns). Fever thought: If we ever want to be mobile, these probably won't be displayed as `columns`. This option suffers from being too visually expressive in general.

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+
+#### `add`
+
+An object of columns to add to the piece type manager. Each column is an object with its own configuration. Column properties include:
+
+| Property | Description |
+| ------- | ------- |
+| `label` | Recommended, but not required, to show a column header label. |
+| `component` | An advanced option to use a custom Vue component for table cells in this column. See core components `AposCellBasic` (default) and `AposCellDate` for examples. |
```
localghost443, line 216: flagging that we should link to this when we can. Is there a master list of these options somewhere? Am I thinking about this wrong?

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+  columns: {
+    // 👇 Hides the column showing when the article was last updated.
+    remove: [ 'updatedAt' ]
+  }
+};
+```
+
+### `instantiate`
+
+Set to `false` to prevent the module from being fully instantiated in the application. The primary purpose of this option is to create a base class module that others will [extend](#extend) but that will not be used directly.
```
localghost443, line 251: Example? Out of my league.

localghost443 (changes requested):

docs/guide/major-changes.md
```diff
@@ -162,7 +162,7 @@ On the other hand, `@apostrophecms/page:beforeSend` will not be listened for as
 
 ### Queries
 
-In modules for piece and page types, the `queries` section replaces the Apostrophe 2.x "cursors" feature. When making a database query with Apostrophe, you can chain together the query "builders" declared in the `builders` sub-section, then finish the job with the query "methods" declared in the `builders` sub-section. Most of the time you'll just need those you inherit from `@apostrophecms/doc-type`, such as `and`, `project`, and `toArray`.
+In modules for piece and page types, the `queries` section replaces the Apostrophe 2.x "cursors" feature. When making a database query with Apostrophe, you can chain together the query "builders" declared in the `builders` subsection, then finish the job with the query "methods" declared in the `builders` subsection. Most of the time you'll just need those you inherit from `@apostrophecms/doc-type`, such as `and`, `project`, and `toArray`.
```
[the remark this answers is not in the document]
abea: We're not. I was simply making our use of "subsection" consistent. And clearly taunting you.

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+| [`improve`](#improve) | String | No | Name a module to update its functionality | All |
+| [`options`](#options) | Object | No | Configure module options | All |
+| [`fields`](#fields) | Object | Yes | Configure doc type fields | Doc, Widget |
+| [`filters`](#filters) | Object | Yes | Configure piece type filters | Piece |
+| [`columns`](#columns) | Object | Yes | Configure piece type manager columns | Piece |
+| [`instantiate`](#instantiate) | Boolean | No | Prevent a module from being fully instantiated | All |
+
+### `extend`
+
+Name a module to use as a base class for your custom module. "Extending" a module this way makes all of the functionality and configuration of the base class available in the new module, other than its `name` property.
```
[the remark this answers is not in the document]
abea: Maybe? You are setting the value as a module name. I'll think on it.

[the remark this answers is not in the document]
abea: I'm trying "Identify"

docs/reference/module-overview.md
```diff
@@ -0,0 +1,284 @@
...
+
+The most common base class modules include:
+
+- `@apostrophecms/piece-type` to add a new custom piece type
```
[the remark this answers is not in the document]
abea: "Custom" here means that you are building it from scratch, rather than configuring a core or installed module. I agree it's not the best term for both parts of this sentence, surely.

docs/reference/module-overview.md
abea: I don't think it should be hyphenated. ["Content type"](https://www.drupal.org/getting-started/6/admin/content/types) is a similar case from our PHP friends, which isn't hyphenated, for example.

docs/reference/module-overview.md
abea: I'm concerned that we're providing too many examples in the reference section already. This is a really deep cut, so I'd rather one day create a recipe/how-to demonstrating a use case. Until then, leave it as "if you don't know what it's for, you probably don't need it."

docs/reference/module-overview.md
```diff
@@ -54,13 +54,11 @@ modules.export = {
...
 ```
 
-You should not use both `improve` and `extend` in a single module.
-
 ### `improve`
 
 Similarly to `extend`, `improve` is used to name another existing module. Instead of _adopting_ the existing module's configuration, using `improve` _changes_ the named module's configuration.
 
-**This is only valid in modules that are built independently and then installed into Apostrophe apps**, either on their own or as a part of a bundle. It is most often used to improve core Apostrophe modules.
+**This is only valid in modules that are installed into an Apostrophe app**, either on their own or as a part of a bundle, and not in those that are built into the app directly. It is most often used to add functionality to core Apostrophe modules.
```
boutell, line 61: If you need to alter the behavior of an installed module in your application-level code, just create an `index.js` file at project level and put your additional code and configuration there. `improve` does much the same thing, but allows you to do it in an npm module where the name can't be the same as the original.

docs/reference/module-overview.md
```diff
@@ -120,13 +122,13 @@ modules.export = {
 
 #### `group`
 
-An object of field groups. The groupings are used by editing interface. It _does not apply to widget modules_, which do not have a tabbed interface.
+An object of field groups. Groupings are used by editing interface. Note that `group` _does not apply to widget modules_.
```
boutell, line 125: "Are used by the"

docs/reference/module-overview.md
```diff
@@ -120,13 +122,13 @@ modules.export = {
...
 
 Groups are added as an object with their name as the object key and the following properties:
-- `label`: The visible label (a string) on the interface tab
+- `label`: The visible label (a string) for the group
 - `fields`: An array of field names to include in the group
 
-The `@apostrophecms/doc-type` module groups the default fields in a `basics` group, which you can override. If overriding the `basics` group, you will need to add those core fields to a group in the module. Any fields not added to a group will be placed in an "Ungrouped" section in the editing interface.
+The `@apostrophecms/doc-type` module arranges the default fields in a `basics` group. You can override the `basics` group, but those default fields will become ungrouped unless you arrange them again. Any fields not added to a group will be placed in an "Ungrouped" section in the editing interface.
```
boutell, line 132: What about utility?

docs/reference/module-overview.md
```diff
@@ -151,19 +153,21 @@ modules.export = {
...
+Filter properties include:
 
 | Property | Description |
 | ------- | ------- |
 | `label` | Recommended, but not required, to customize the human-readable label. |
-| `inputType` | Choose an input field type for the filter from the following: `radio`, `checkbox`, or `select`. Defaults to `select`. |
-| `choices` | Manually set an array of choices instead of letting Apostrophe find the valid choices from the database. Choices include `label` and `value` properties. |
+| `inputType` | Set the an input field type for the filter. Options include `radio`, `checkbox`, or `select`. Defaults to `select`. |
+| `choices` | Manually set an array of choices. Choices require `label` and `value` properties. |
 | `def` | The default value for the manager filter. |
```
boutell, line 172: What about `remove`?

boutell (changes requested):

docs/guide/major-changes.md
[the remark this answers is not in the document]
localghost443: Lol ok I won’t dig in

docs/reference/module-overview.md
```diff
@@ -54,13 +54,11 @@ modules.export = {
 };
 ```
 
-You should not use both `improve` and `extend` in a single module.
-
 ### `improve`
 
 Similarly to `extend`, `improve` is used to name another existing module. Instead of _adopting_ the existing module's configuration, using `improve` _changes_ the named module's configuration.
```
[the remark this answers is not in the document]
abea: 💯

docs/reference/module-overview.md
```diff
@@ -6,24 +6,24 @@ Module configuration objects may use the following configuration properties.
 
 | Setting name | Value type | [Cascades](#cascading-settings) | Description | Module types |
 | ------- | ------- | ------- | ------- | :-----: |
-| [`extend`](#extend) | String | No | Name a base class module | All |
-| [`improve`](#improve) | String | No | Name a module to update its functionality | All |
+| [`extend`](#extend) | String | No | Identify the base class module | All |
+| [`improve`](#improve) | String | No | Identify a module to update its functionality | All |
```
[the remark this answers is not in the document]
abea: I like "enhance." I'll leave the "doesn't create a new one" for the description below.

docs/reference/module-overview.md
```diff
@@ -73,13 +71,17 @@ modules.export = {
 };
 ```
 
-You should not use both `improve` and `extend` in a single module.
+You should not use both `improve` and `extend` in a single module. If "improving" an existing module, that existing module has already taken care of the "extending."
```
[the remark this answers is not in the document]
abea: Sure, but it's still "extending." I'm (I think) intentionally not explicitly saying that it will use `extend`, but it's still doing the thing implicitly.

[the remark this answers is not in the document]
abea: I'll not in `extend` that leaving it off implicitly extends the base module.

docs/reference/module-overview.md
abea: ... off limits. Just kidding, but maybe we shouldn't allow utility fields to be moved? Or to be ungrouped? @localghost443 ?

docs/reference/module-overview.md
abea: It's the next subsection.

docs/reference/module-overview.md
boutell: ^^^ I think the above is important.

docs/reference/module-overview.md
boutell: I think it's best to leave it up to re-grouping like anything else, since we can't anticipate all future situations and our extensibility is here to accommodate them, but it's fair to say it would be best practice to leave the standard utility fields in that group when they exist.

boutell (changes requested):

docs/reference/module-overview.md
abea: I added that in a note right after this.

docs/guide/major-changes.md
```diff
@@ -162,7 +162,7 @@ On the other hand, `@apostrophecms/page:beforeSend` will not be listened for as
 
 ### Queries
 
-In modules for piece and page types, the `queries` section replaces the Apostrophe 2.x "cursors" feature. When making a database query with Apostrophe, you can chain together the query "builders" declared in the `builders` subsection, then finish the job with the query "methods" declared in the `builders` subsection. Most of the time you'll just need those you inherit from `@apostrophecms/doc-type`, such as `and`, `project`, and `toArray`.
+In modules for piece and page types, the `queries` section replaces the Apostrophe 2.x "cursors" feature. When making a database query with Apostrophe, you can chain together the query "builders" declared in the `builders` subsection, then finish the job with the query "methods" declared in the `methods` subsection. Most of the time you'll just need those you inherit from `@apostrophecms/doc-type`, such as `and`, `project`, and `toArray`.
```
boutell, line 165: ... As well as the query builders that are added automatically for most non-area fields in your schema. (In the "most of the time you'll just need" section)

boutell (changes requested):

docs/guide/major-changes.md
abea: This was simply fixing a typo. We'll be totally rewriting this page later.

docs/reference/module-overview.md
localghost443: I tend to favor the idea hat utility fields shouldn't be moved. I don't think it offers any utility and adds complexity. Its definitely best practice not to move them. Some powers shouldn't be given or encouraged.

localghost443 (approved):

docs/reference/module-overview.md
```diff
@@ -0,0 +1,291 @@
...
+| [`improve`](#improve) | String | No | Identify a module to enhance | All |
+| [`options`](#options) | Object | No | Configure module options | All |
+| [`instantiate`](#instantiate) | Boolean | No | Prevent a module from being fully instantiated | All |
+| [`fields`](#fields) | Object | Yes | Configure doc type fields | Doc, Widget |
+| [`filters`](#filters) | Object | Yes | Configure piece type filters | Piece |
+| [`columns`](#columns) | Object | Yes | Configure piece type manager columns | Piece |
+
+### `extend`
+
+Identify a module to use as a base class for your custom module. "Extending" a module this way makes all of the functionality and configuration of the base class available in the new module, other than its `name` property.
```
localghost443, line 19: "custom module" -- not going to stop this from being 'finished', but flagging this again. Custom implies aberration, and everything....is a module. So I think this lacks a lot of clarity. We're not using any real, relevant descriptors here, imo

boutell (approved):

docs/reference/module-overview.md
boutell: I think custom implies "your own." But it could just say "a base class for this module."

docs/reference/module-overview.md
abea: Seems like a good idea to me. That should be a new ticket for "some day"

PR #45, opened 2021-03-12
Second group of module config sections

abea:


docs/reference/module-api/module-overview.md
```diff
@@ -257,14 +257,231 @@ modules.export = {
 };
 ```
 
-## Customization functions
+## Initialization function
 
 ### `async init(self)`
-### `beforeSuperClass(self)`
+
+This function runs once when the Apostrophe app first start up. It takes the module, as `self`, as an argument.
```
boutell, line 264: first starts up.

boutell: Based on years and years of developer emails and questions, I think it's worth adding "it does NOT run on every request. For that, see event handlers, specifically `@apostrophecms/page:beforeSend`."

boutell: Many new Apostrophe developers don't really get that node servers are long-lasting processes.

docs/reference/module-api/module-overview.md
```diff
@@ -257,14 +257,231 @@ modules.export = {
...
+```javascript
+// modules/product/index.js
+module.exports = {
+  // ...
+  options: {
+    alias: 'product',
+    themeColor: 'blue'
+  },
+  init(self) {
+    // 👇 Making `self.options.themeColor` available on `apos.product.theme`.
```
boutell, line 277: Nice.

docs/reference/module-api/module-overview.md
```diff
@@ -257,14 +257,231 @@ modules.export = {
...
+    self.theme = self.options.themeColor;
+
+    // 👇 Adding a data migration related to this module using a method from a
+    // separate module.
+    self.apos.migration.add('blurb', async () => {
+      await self.apos.migration.eachDoc({
+        type: 'product'
+      }, async (doc) => {
+        if ((doc.category) === 'old-category') {
+          doc.category = 'new-category';
```
boutell, line 287: Setting doc.category in this way has no effect on the database, only what is in updateOne matters, so this should become part of the `$set` object.

docs/reference/module-api/module-overview.md
```diff
@@ -257,14 +257,231 @@ modules.export = {
...
+### The extension pattern
+
+Several of these sections use an extention pattern (the sections prefixed with "extend"). These sections are used to add functionality to matching features of the module's base class. Functions in `extendMethods` enhance identically named functions in the base class' `methods` section, `extendApiRoutes` enhances API routes from the base class' `apiRoutes`, and so on.
+
+Each individual function included in the extension section takes a `_super` argument in addition to the same arguments as the original function. `_super` is the original function, which should be called within the new extension.
+
+If a piece type included the `insert` in its `extendMethods` section to alter the piece titles, it might look like this:
+
+```javascript
+  enhanceMethods(self) {
```
boutell, line 336: `extendMethods`, not `enhanceMethods`.

boutell (changes requested):
Great stuff, minor things.

docs/reference/module-api/module-overview.md
abea: Good point. This isn't specifically about pages, so I don't think we should point to that specifically, but we can link over to an event handlers guide.

abea: Adding "To run code on every request, or in other situations when the app is running, see the event handlers documentation." Let me know if there's other language for that you think would be more appropriate.

docs/reference/module-api/module-overview.md
abea: Yep. Lazy code copying.

docs/reference/module-api/module-overview.md
```diff
@@ -257,14 +257,231 @@ modules.export = {
...
+        piece.title = `🆕 ${piece.title}`;
+
+        _super(req, piece, options);
+      }
+    }
+  }
+```
+
+::: warning
+Extension functions should *always* use the `_super` argument to call the original function. If you want to *completely* overwrite the inherited function, instead add a matching function in the section without the "extend" prefix. For example, we could completely overwrite the `insert` method in our piece type by including our own `insert` function in the `methods` section.
```
localghost443, line 348: "instead add a matching function in the section without the "extend" prefix" -> "add a matching function in the section without the "extend" prefix instead".

docs/reference/module-api/module-overview.md
```diff
@@ -257,14 +257,231 @@ modules.export = {
...
+        return piece;
+      }
+    };
+  }
+};
+```
+
 ### `components(self)`
+
+This function section returns an object containing functions that power asynchronous template components. These template components allow for asynchronous data requests in normally synchronous template rendering.
```
localghost443, line 411: "This function section" doesn't seem necessary. "These template components" -> "asynchronous template component".

boutell (approved):

localghost443 (approved):

PR #46, opened 2021-03-15
Module overview 3

abea:
Yet more module overview sections

- helpers
- extendHelpers
- restApiRoutes
- extendRestApiRoutes
- apiRoutes
- extendApiRoutes
- renderRoutes
- routes
- handlers
- extendHandlers

We might want to do some reordering later, but I didn't want to make it more confusing to review.


docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
...
+    };
+  }
+};
+```
+
 ### `routes(self)`
+
+Add standard Express routes. The `routes` function takes takes the module as an argument and must return an object with properties for the relevant [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods), including `get`, `post`, `patch`, and `delete`. Each of those properties should be set to an object of functions.
+
+Each `routes` function takes the Express arguments `req` (the [request object](https://expressjs.com/en/api.html#req)) and `res` (the [response object](https://expressjs.com/en/api.html#res)). The functions must return using the `res.redirect` or `res.send` methods.
```
boutell, line 746: How about: "the functions must generate a response via `res` to avoid leaking resources, typically using the `res.redirect` or `res.send` methods." There are other, less often used ways to end a response, such as calling `res.send`, so I don't want to make it sound like Apostrophe is doing anything other than letting you do as you like with Express here.

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
...
+  }
+};
+```
+
 ### `handlers(self)`
+
+The `handlers` function takes the module as an argument and must return an object. The object keys should be names of existing server-side events. The value of those event keys should be an object of functions to execute when those events fire.
+<!-- TODO: Link to the reference to or guide on server-side events when available. -->
+
+Events belonging to the same module where the handlers are defined, or from its base class, can be referenced simply by name, e.g., `beforeInsert` for any piece type. You may also add handlers in one module that respond to events in other modules. Those event names should be prefixed with the name of the module where the event fires followed by a colon, e.g., `@apostrophecms/page:beforeSend`.
```
boutell, line 774: Event handlers may be async functions, and they will be awaited before execution continues.

boutell (changes requested):
Small stuffs. This is so 💯

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
 ```
 
 ### `helpers(self)`
+
+The `helpers` section returns an object of functions that add template utility methods. The individual helper methods may take any arguments that you plan to pass them in templates.
```
localghost443, line 487: "The `helpers` section" could probably be dropped. We're already there.

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
...
+        return `${price} 🤑`;
+      }
+    };
+  }
+};
+```
+
 ### `restApiRoutes(self)`
+
+Apostrophe includes a full REST API for [pieces](/reference/api/pieces.md) and [pages](/reference/api/pages.md). Those route handlers can be totally overridden by adding identically named functions to object returned by the `restApiRoutes` function. If you simply wish to add to the existing behavior of the REST API routes, see [`extendRestApiRoutes`](#extendrestapiroutes-self).
```
localghost443, line 543: "**Those route handlers can be totally overridden** by adding identically named functions **to object** returned by the `restApiRoutes` function. If you **simply wish to add to** the existing behavior of the REST API routes, see [`extendRestApiRoutes`](#extendrestapiroutes-self)."

"Those route handlers can be totally overridden" -> "You can completely override these route handlers" (completely optional?)

"to object" -> "to objects"

"If you **simply wish to add to**" -> "To add to and extend existing behavior"

docs/reference/module-api/module-overview.md
boutell: `routes` should be used only where `apiRoutes` or `restApiRoutes` or `renderRoutes` won't do, such as when you need to use `res.redirect` or pipe a stream. We should state this clearly to help guide people to the painless solution. Failing to do the right thing with `res` or handle errors etc. is a common cause of Node.js bugs, which is why apiRoutes exists.

(At some point I'd like to add a way to use `res.redirect` via a return value. Piping a stream will probably always be an "uh... just use routes()" scenario.")

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
...
+        };
+      }
+    };
+  }
+};
+```
+
 #### `extendRestApiRoutes(self)`
+
+Extend the behavior of existing REST API routes in the `extendRestApiRoutes` section. This function must return an object of functions. See [`restApiRoutes`](#restapiroutes-self) for the valid function names.
```
localghost443, line 576: "in the `extendRestApiRoutes` section" -> remove

docs/reference/module-api/module-overview.md
boutell, line 543: You imply it, but this really needs to be emphasized:

All piece type modules, as well as the `@apostrophecms/page` module, already have a complete REST API and you should not use `restApiRoutes` with these. The most likely use of `restApiRoutes` in a project would be to provide a custom REST API to a database or service not already built into Apostrophe. `restApiRoutes` is also not for custom routes that don't map to one of the standard REST URLs. If you need to add a custom route over and above the REST API of pieces or pages, you should do that with `apiRoutes`.

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
...
+        }
+      }
+    };
+  }
+};
+```
+
+#### Naming routes
+
+If route properties do *not* begin with a forward slash, they can be reached via the pattern: `/api/v1/` followed by the module name, a forward slash, and the API route handler's property name, e.g., `/api/v1/product/cheapest` for the `cheapest` route on the `product` module.
```
localghost443, line 636: `cheapest` could probably be replaced by something that translates a bit better, since its pretty much slang. Using product....my first thought is `brand/s` or a category like `denim-jacket/s`. Denim and jacket are both fairly universal, while being very American!

Bonus points if we find a way to use infant bodysuit

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
...
+| `'locked'` | 409 |
+| `'unprocessable'` | 422 |
+| `'unimplemented'` | 501 |
+
+Passing a different value as the first argument to `self.apos.error()` will set the response code to 500.
+<!-- TODO: Link to the method's own documentation page when available for more. -->
+
+#### `extendApiRoutes(self)`
+
+Extend the behavior of existing API routes (set in `apiRoutes`) in the `extendApiRoutes` section. This function must return an object as described in [`apiRoutes`](#apiroutes-self).
```
localghost443, line 683: "in the `extendApiRoutes` section" -> dont need (imo)

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
...
+        }
+      }
+    };
+  }
+};
+```
+
 ### `renderRoutes(self)`
+
+Add custom API routes to return rendered templates. The `apiRoutes` function takes takes the module as an argument and must return an object with properties for the relevant [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods), including `get`, `post`, `patch`, and `delete`. Each of those properties should be set to an object of functions.
```
localghost443, line 711: "the module as an argument" -> is this the module name?

docs/reference/module-api/module-overview.md
localghost443, line 774: "simply by name" -> "by name"

docs/reference/module-api/module-overview.md
localghost443: I love wishes tho

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,354 @@ module.exports = {
...
+| `'forbidden'` | 403 |
+| `'notfound'` | 404 |
+| `'required'` | 422 |
+| `'conflict'` | 409 |
+| `'locked'` | 409 |
+| `'unprocessable'` | 422 |
+| `'unimplemented'` | 501 |
+
+Passing a different value as the first argument to `self.apos.error()` will set the response code to 500.
+<!-- TODO: Link to the method's own documentation page when available for more. -->
```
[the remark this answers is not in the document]
abea: I'm holding off on full documentation of that here since it could be a big chunk. I'm going to link over to it once available instead. There'll be a sentence like "See more information on using the `error()` method."

docs/reference/module-api/module-overview.md
abea: I've added a new :::tip block a few lines down.

docs/reference/module-api/module-overview.md
abea: I've been internally debating how much to explicitly say that things can be async. It's probably a good idea to do more.

docs/reference/module-api/module-overview.md
abea: > The most likely use of restApiRoutes in a project of your own would be to provide a custom REST API to a database or service not already built into Apostrophe.

@boutell Could you suggest a more specific use case? I want to make sure I'm communicating what you're saying correctly. Like I create a `rando` module that is not a piece, then I add a REST API that actually hits a third-party database?

abea: Maybe `apiRoutes` should appear first in the page.

abea: I'm redoing that whole section based on Tom's notes.

docs/reference/module-api/module-overview.md
abea: These should say "function" not "section."

abea: Or nothing.

docs/reference/module-api/module-overview.md
abea: 😕 "cheapest" simply means "most cheap." Least expensive.

docs/reference/module-api/module-overview.md
abea: No, it's the actual module. That's what `self` is.

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,366 @@ module.exports = {
...
 ### `routes(self)`
+
+Add standard Express routes. The `routes` function takes takes the module as an argument and must return an object with properties for the relevant [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods), including `get`, `post`, `patch`, and `delete`. Each of those properties should be set to an object of functions. Route functions may be asynchronous (async).
+
+Each `routes` function takes the Express arguments `req` (the [request object](https://expressjs.com/en/api.html#req)) and `res` (the [response object](https://expressjs.com/en/api.html#res)). The functions must generate a response via `res` to avoid leaking resources, typically using the `res.redirect` or `res.send` methods.
+
+See [Naming routes](#naming-routes) for more on function names and their route URLs.
+
+::: tip
+We recommend using `apiRoutes` or `restApiRoutes` whenever possible before using `routes`. They are designed to be easier to use and handle the potential pitfalls of Express routes. There are situation when writing Express routes may be necessary, such as when you need to use `res.redirect` or pipe a stream.
```
localghost443, line 759:
```suggestion
We recommend using `apiRoutes` or `restApiRoutes` whenever possible before using `routes`, as they handle potential pitfalls of Express routes. There are situations where writing Express routes may be necessary, such as when you need to use `res.redirect` or pipe a stream.
```

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,366 @@ module.exports = {
...
+  }
+};
+```
+
 ### `handlers(self)`
+
+The `handlers` function takes the module as an argument and must return an object. The object keys should be names of existing server-side events. The value of those event keys should be an object of functions to execute when those events fire. Event handlers may be asynchronous (async) functions.
+<!-- TODO: Link to the reference to or guide on server-side events when available. -->
+
+Events belonging to the same module where the handlers are defined, or from its base class, can be referenced by name, e.g., `beforeInsert` for any piece type. You may also add handlers in one module that respond to events in other modules. Those event names should be prefixed with the name of the module where the event fires followed by a colon, e.g., `@apostrophecms/page:beforeSend`.
```
localghost443, line 786:
```suggestion
Events belonging to the same module where the handlers are defined, or from its base class, can be referenced by name, e.g., `beforeInsert` for any piece type. You may also add `handlers` in one module that respond to events in other modules. Those event names should be prefixed with the name of the module where the event occurs, followed by a colon, e.g., `@apostrophecms/page:beforeSend`.
```

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,366 @@ module.exports = {
 ```
 
 ### `helpers(self)`
+
+`helpers` returns an object of functions that add template utility methods. The individual helper methods may take any arguments that you plan to pass them in templates. Helper functions must run synchronously.
```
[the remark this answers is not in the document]
boutell: How about: "the individual helper methods may accept arguments when called from your templates."

docs/reference/module-api/module-overview.md
```diff
@@ -483,18 +483,366 @@ module.exports = {
...
+| `'unimplemented'` | 501 |
+
+Passing a different value as the first argument to `self.apos.error()` will set the response code to 500.
+<!-- TODO: Link to the method's own documentation page when available for more. -->
+
+#### `extendApiRoutes(self)`
+
+Extend the behavior of existing API routes (set in `apiRoutes`) in `extendApiRoutes`. This function must return an object as described in [`apiRoutes`](#apiroutes-self).
+
+Each extended API route function should accept the original function as `_super` and the `req` request object. They should return data in a similar format to the existing API route.
```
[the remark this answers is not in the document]
boutell: "Your extended API route function should"?

boutell (approved):
I'm approving with suggestions on Ben's grammar notes.

docs/reference/module-api/module-overview.md
abea: "handlers" is also the name of the type of function, like we call then "Event handlers" earlier.

localghost443 (approved):

PR #47, opened 2021-03-16
Module overview 4

abea:
Adds the query builders section. A lot going on there. We'll be linking out to a separate page with the core query builders and query methods, as well as a guide on how to use this stuff like a human.


docs/reference/module-api/module-overview.md
```diff
@@ -838,19 +838,82 @@ module.exports = {
...
-### `queries(self)`
+### `queries(self, query)`
+
+The `queries` function is part of the Apostrophe API for querying the database. It takes two arguments: the module (`self`) and the query that will be called (`query`). It must return an object. That object can have two properties:
+<!-- TODO: Link to a real guide on using queries when available. This is simply reference. -->
+
+| `queries` properties | Description |
+| ---- | ---- |
+| [`builders`](#builders) | An object of "builders," or chainable methods, that refine the query or otherwise change its behavior prior to execution. |
+| [`methods`](#methods) | An object of method functions that are used to execute some action on the query. |
```
localghost443, line 849:
```suggestion
| [`methods`](#methods) | An object of method functions that execute some action on the query. |
```
I think we could ditch "that are used". Might be wrong here. I'm tripping over "An object" "that are".

docs/reference/module-api/module-overview.md
```diff
@@ -838,19 +838,82 @@ module.exports = {
...
+
+#### `builders`
+
+Query builders are defined as objects with a set of specific properties available to them. Builders often take an argument or have a default value that is used instead.
```
localghost443, line 853:
```suggestion
Query builders are defined as objects with a set of properties available to them. Builders often take an argument or use a default value.
```
I think we could remove "specific" and "is used instead" can be simply "or".

docs/reference/module-api/module-overview.md
```diff
@@ -838,19 +838,82 @@ module.exports = {
...
+<!-- TODO: Link to a reference of core builders when available. -->
+
+| Builder properties | Description |
+| ---- | ---- |
+| `def` | The default value for the builder. |
+| `launder` | A function used to validate values passed as arguments when `applyBuildersSafely` is called on the query. Returns `true` if valid.<br />This is required to use the builder in a REST API query string. |
+| `choices` | A function returning an array of "choice" objects with `value` and `label` properties. This is returned from the `toChoices` query method if set. |
+| `set` | A function to called when the builder is invoked, instead of the default `query.set` method. This should usually include running the `query.set` itself in addition to other work. |
```
localghost443, line 861:
```suggestion
| `set` | A function called when the builder is invoked instead of the default `query.set` method. This usually includes running the `query.set` itself. |
```
I remove "to", assuming the original intent was "to call". I removed "in addition to other work" because it implied by "should usually include", which is in turn "usually includes"

docs/reference/module-api/module-overview.md
```diff
@@ -838,19 +838,82 @@ module.exports = {
...
+| `prefinalize` | A function to run before any builder `finalize` steps. Used to alter the query with other builders. |
```
localghost443, line 862: lol

docs/reference/module-api/module-overview.md
```diff
@@ -838,19 +838,82 @@ module.exports = {
...
+| `finalize` | A function to run at the end of the query building phase, prior to being processed by the database. Used to alter the query with other builders. |
+| `after` | A function run to mutate an array of queried items passed in as an argument. This should check for validity prior to mutating. |
```
localghost443, line 864: "This should check" are we staying that the function **will** check for validity? Or that it will if able? Or that the developer should do this check?

docs/reference/module-api/module-overview.md
```diff
@@ -838,19 +838,82 @@ module.exports = {
...
+        }
+      }
+    };
+  }
+};
+```
+
+#### `methods`
+
+An object of methods that can be used to execute queries after any builders have been applied. These functions should use existing query methods or [MongoDB cursor methods](https://docs.mongodb.com/manual/reference/method/js-cursor/) to return documents.
```
localghost443, line 908: "can be used" similar to above. Its less trippy here, though because "Can".

docs/reference/module-api/module-overview.md
abea: The developer should do the check.

localghost443 (approved):

docs/reference/module-api/module-overview.md
```diff
@@ -838,19 +838,82 @@ module.exports = {
...
+
+| Builder properties | Description |
+| ---- | ---- |
+| `def` | The default value for the builder. |
+| `launder` | A function used to validate values passed as arguments when `applyBuildersSafely` is called on the query. Returns `true` if valid.<br />This is required to use the builder in a REST API query string. |
+| `choices` | A function returning an array of "choice" objects with `value` and `label` properties. This is returned from the `toChoices` query method if set. |
+| `set` | A function called when the builder is invoked instead of the default `query.set` method. The `set` function should include running the `query.set` after other work. |
+| `prefinalize` | A function to run before any builder `finalize` steps. Used to alter the query with other builders. |
+| `finalize` | A function to run at the end of the query building phase, prior to being processed by the database. Used to alter the query with other builders. |
+| `after` | A function run to mutate an array of queried items passed in as an argument. The `after` function should include checking for validity prior to mutating. |
```
boutell, line 864: I don't understand where this came from: "The `after` function should include checking for validity prior to mutating."

`after` takes stuff that came back from mongo and does extra work like loading widgets or relationships.

docs/reference/module-api/module-overview.md
```diff
@@ -838,19 +838,82 @@ module.exports = {
 };
 ```
 
-### `queries(self)`
+### `queries(self, query)`
+
+The `queries` function is part of the Apostrophe API for querying the database. It takes two arguments: the module (`self`) and the query that will be called (`query`). It must return an object. That object can have two properties:
```
boutell, line 843: "The query that will be called" is a little unsatisfying but, well, yes, and it avoids language like "the query that is being constructed."

Users are going to read this and bounce pretty hard, thinking "so... what? I call self.queries to make a query?"

But it's reference, not the guide you promise elsewhere, so.

boutell (approved):

docs/reference/module-api/module-overview.md
abea: I mean that you should check that the builder was used.

```
        relationships: {
          def: true,
          async after(results) {
            if (!query.get('relationships')) {
              return;
            }`
```

docs/reference/module-api/module-overview.md
abea: I don't hate "the query that is being constructed"

docs/reference/module-api/module-overview.md
boutell: OK. "validity" implies something more I think. How about "the after function should use `query.get` to confirm that the work should be done."

docs/reference/module-api/module-overview.md
boutell: Cool, I think that would be good language.

docs/reference/module-api/module-overview.md
```diff
@@ -861,7 +861,7 @@ Query builders are defined as objects with a set of properties available to them
 | `set` | A function called when the builder is invoked instead of the default `query.set` method. The `set` function should include running the `query.set` after other work. |
 | `prefinalize` | A function to run before any builder `finalize` steps. Used to alter the query with other builders. |
 | `finalize` | A function to run at the end of the query building phase, prior to being processed by the database. Used to alter the query with other builders. |
-| `after` | A function run to mutate an array of queried items passed in as an argument. The `after` function should include checking for validity prior to mutating. |
+| `after` | A function run to mutate an array of queried items passed in as an argument. The `after` function should return out immediately if the builder was not used on the query. |
```
boutell, line 864: Ah I see you addressed my suggestion here.

docs/reference/module-api/module-overview.md
```diff
@@ -840,7 +840,7 @@ module.exports = {
 
 ### `queries(self, query)`
 
-The `queries` function is part of the Apostrophe API for querying the database. It takes two arguments: the module (`self`) and the query that will be called (`query`). It must return an object. That object can have two properties:
+The `queries` function is part of the Apostrophe API for querying the database. It takes two arguments: the module (`self`) and the query that is being constructed (`query`). It must return an object. That object can have two properties:
```
boutell, line 843: I think this first sentence is going to confuse people a lot and lead to a lot of head-bashing:

"The `queries` function is part of the Apostrophe API for querying the database."

This strongly implies you can call it, and you can't.

I think this would be better:

"The `queries` function of your module is called once when Apostrophe starts up to register custom query builders and methods."

boutell (changes requested):

boutell (approved):

PR #48, opened 2021-03-17
Module overview 5

abea:
The last??? module overview sections.

### DIFF
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -344,7 +344,7 @@ Several of these sections use an extension pattern. The sections prefixed with `
 
 Each function included in an extension section takes the function it is extending as the first argument (`_super`, below) in addition to the original arguments.
 
-If a piece type included the `generate` method in its `extendMethods` section to add a price upon generating placeholder docs, it might look like this:
+The following example updates the `generate` method in the `extendMethods` section to add a placeholder `price` upon creation.
 
 ```javascript
 // modules/product/index.js
@@ -355,19 +355,19 @@ module.exports = {
       // The original `generate` function only takes `index` as an argument.
       generate(_super, index) {
         // Using _super with the original argument to generate a sample piece.
-        const piece = _super(index);
+        const product = _super(index);
         // Adding additional functionality.
-        piece.price = Math.random() * 100;
-        // Returning the generated piece, exactly as the original `generate`
-        // method does.
-        return piece;
+        product.price = Math.random() * 100;
+        // Returning the generated product piece, exactly as the original
+        // `generate` method does.
+        return product;
       }
     };
   }
 };
 ```
 
-Or this, if extending a [REST API function](#restapiroutes-self):
+This example shows another example, extending a [REST API function](#restapiroutes-self):
 
 ```javascript
 // modules/product/index.js
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -1,9 +1,9 @@
 # Module properties
 
 Module configuration objects may use the following configuration properties. The overall categories are broadly defined:
-- [Configuration settings](#configuration-settings): Static module settings. They generally cannot make use of other module settings.
-- [Initialization function](#initialization-function): A function that runs once when the Apostrophe app first starts up.
-- [Customization functions](#customization-functions): Settings via functions that at least have access to the module itself as an argument. They often make use of other settings.
+- [Configuration settings](#configuration-settings): Static module settings. These generally cannot use other module settings.
+- [Initialization function](#initialization-function): A function that runs once during application startup.
+- [Customization functions](#customization-functions): Settings via functions that have access to the module itself as an argument and can access other settings.
 
 ## Configuration settings
 
@@ -20,7 +20,7 @@ Module configuration objects may use the following configuration properties. The
 
 ### "Cascading" settings
 
-Many Apostrophe module sections are structured as objects with `add`, `remove` and (sometimes), `group` properties. This pattern allows these settings to "cascade" from the base classes through to project level classes. The subsections help Apostrophe manage inherited options without requiring those inherited options to be re-declared by developers.
+Many Apostrophe module sections are structured as objects with `add`, `remove`, and sometimes `group` properties. This pattern allows these settings to "cascade" from the base classes through to project level classes. The subsections help Apostrophe manage inherited settings without requiring those inherited settings be re-declared by developers.
 
 If new setting options are needed, they go in `add`. If base class setting options are no longer wanted, they go in `remove`. And in some cases where grouping happens in the interface, they can be organized in `group`.
 
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -322,7 +322,7 @@ Each of these function sections takes the module, as `self`, as an argument. Thi
 | [`queries`](#queries-self-query) | Add database query methods |
 | [`extendQueries`](#extendqueries-self) | Extend base class database query methods |
 | [`middleware`](#middleware-self) | Add standard Express middleware to be called on *every* request. |
-| [`tasks`](#tasks-self) | REPLACE-ME |
+| [`tasks`](#tasks-self) | Add task functions that can be run from the command-line. |
 
 ### The extension pattern
 
@@ -903,6 +903,60 @@ module.exports = {
 ```
 
 ### `tasks(self)`
+
+`tasks` takes the module as an argument and returns an object of functions that define command-line tasks. Task properties include:
+
+
+| Property | Description |
+| ------- | ------- |
+| `usage` | A string describing the task and how to use it. It is printed on the command-line. |
+| `task` | The task function. It maybe asynchronous. |
+| `afterModuleInit` | Set to `true` to run the task after modules are initiated but *before* they are fully "awake".
+| `exitAfter` | Only relevant if `afterModuleInit` is `true`. Set to `false` to *avoid* exiting the Apostrophe process on completion. Uncommon. |
+
+Task functions takes the object `argv` as an argument, which includes the arguments passed after the task command. As documented by the [Boring](https://www.npmjs.com/package/boring) utility:
+
+> When someone types:
+>
+> ```
+> node app jump sideways --foo --bar=whee --super-cool=totally
+> ```
+>
+> You get:
+>
+> ```
+> {
+>   _: [ "jump", "sideways"],
+>   foo: true,
+>   bar: "whee",
+>   "super-cool": "totally"
+> }
+> ```
+
+```javascript
+// modules/product/index.js
+module.exports = {
+  // ...
+  tasks(self, options) {
+    return {
+      // Since the module is named product, then you can run this CLI task by
+      // typing: `node app product:list` in the CLI.
+      list: {
+        usage: 'List the titles of each product.',
+        async task(argv) {
+          // Get an req object with admin privileges. You can also use getAnonReq.
+          const req = self.apos.task.getReq();
+          const pieces = await self.find(req).toArray();
+
+          for (const piece of pieces) {
+            console.log(piece.title);
+          }
+        }
+      }
+    }
+  }
+};
+```
 
 ## "Cascading" settings
 
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -435,7 +435,7 @@ To maintain the same application, they should return the same type of response a
 
 ### `components(self)`
 
-This function section returns an object containing functions that power asynchronous template components. These template components allow for asynchronous data requests in normally synchronous template rendering.
+Returns an object containing functions that power asynchronous template components. These template components allow for asynchronous data requests in normally synchronous template rendering.
 
 Each template component function should take the arguments:
 
@@ -861,15 +861,15 @@ module.exports = {
 
 Extend the behavior of existing event handlers (set in `queries`) in the `extendQueries` section. This function must return an object as described in `queries`.
 
-Each extended query builder or method should accept the original function as `_super` followed by its original arguments, if there are any. Extended query builders and methods will be matched with the base class builder or method using the same name. Methods should return data in a similar format to the existing API route.
+Each extended query builder or method should accept the original function as `_super` followed by its original arguments. Extended query builders and methods will be matched with the base class builder or method using the same name. Methods should return data in a similar format to the existing API route.
 
 ### `middleware(self)`
 
-Add standard Express middleware to be called on *every* request. The `middleware` function takes the module as an argument and must return an object of [middleware functions](https://expressjs.com/en/guide/using-middleware.html). This is a good place to import third-party middleware into Apostrophe.
+Add standard Express middleware to be called on *every* request. The `middleware` function takes the module as an argument and must return an object of [middleware functions](https://expressjs.com/en/guide/using-middleware.html). This is a good place to import third-party middleware.
 
-Note that it can often be simpler to add an event handler or `await` a method in an API route instead.
+Note that it is often better to add an event handler or `await` a method in an API route instead.
 
-`ours` could also be an object with a `before` property and a `middleware` property, in which case it would run `before` the middleware of the module with the specified name
+`checkIp` could also be an object with a `before` property and a `middleware` property, in which case it would run `before` the middleware of the module with the specified name
 
 ```javascript
 // modules/limiter/index.js
@@ -877,7 +877,7 @@ module.exports = {
   // ...
   middleware(self, options) {
     return {
-      ours(req, res, next) {
+      checkIp(req, res, next) {
         // Restrict access by IP address, in a crude way.
         const allowlist = [ '127.0.0.1', '::1' ];
 
@@ -899,7 +899,7 @@ module.exports = {
   // ...
   middleware(self, options) {
     return {
-      ours: {
+      checkIp: {
         // 👇 Same as above, but with `before` and `middleware` properties.
         before: '@apostrophecms/login',
         middleware: function (req, res, next) {
@@ -919,14 +919,14 @@ module.exports = {
 
 ### `tasks(self)`
 
-`tasks` takes the module as an argument and returns an object of functions that define command-line tasks. Task properties include:
+`tasks` takes the module as an argument and returns an object of functions that define command line tasks. Task properties include:
 
 
 | Property | Description |
 | ------- | ------- |
-| `usage` | A string describing the task and how to use it. It is printed on the command-line. |
-| `task` | The task function. It maybe asynchronous. |
-| `afterModuleInit` | Set to `true` to run the task after modules are initiated but *before* they are fully "awake".
+| `usage` | A string describing the task and how to use it. It is printed on the command line. |
+| `task` | The task function. Can be asynchronous. |
+| `afterModuleInit` | Set to `true` to run the task after modules are initiated but *before* they are fully active.
 | `exitAfter` | Only relevant if `afterModuleInit` is `true`. Set to `false` to *avoid* exiting the Apostrophe process on completion. Uncommon. |
 
 Task functions takes the object `argv` as an argument, which includes the arguments passed after the task command. As documented by the [Boring](https://www.npmjs.com/package/boring) utility:
@@ -954,8 +954,8 @@ module.exports = {
   // ...
   tasks(self, options) {
     return {
-      // Since the module is named product, then you can run this CLI task by
-      // typing: `node app product:list` in the CLI.
+      // Since the module is named product, then you can run this command
+      // line task by typing: `node app product:list` in the CLI.
       list: {
         usage: 'List the titles of each product.',
         async task(argv) {
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -863,7 +863,7 @@ Each extended query builder or method should accept the original function as `_s
 
 Add standard Express middleware to be called on *every* request. The `middleware` function takes the module as an argument and must return an object of [middleware functions](https://expressjs.com/en/guide/using-middleware.html). This is a good place to import third-party middleware into Apostrophe.
 
-Note that it can often be simpler to add an event handler for many purposes instead.
+Note that it can often be simpler to add an event handler or `await` a method in an API route instead.
 
 `ours` could also be an object with a `before` property and a `middleware` property, in which case it would run `before` the middleware of the module with the specified name
 
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -930,14 +930,14 @@ Task functions takes the object `argv` as an argument, which includes the argume
 > Input:
 >
 > ```
-> node app my-module:runIt taskOption --foo --bar=baz --use-color=green
+> node app my-module:run-it taskOption --foo --bar=baz --use-color=green
 > ```
 >
 > Response:
 >
 > ```
 > {
->   _: [ "runIt", "taskOption"],
+>   _: [ "my-module:run-it", "taskOption"],
 >   foo: true,
 >   bar: "baz",
 >   "use-color": "green"
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -1,6 +1,9 @@
 # Module properties
 
-Module configuration objects may use the following configuration properties.
+Module configuration objects may use the following configuration properties. The overall categories are broadly defined:
+- [Configuration settings](#configuration-settings): Static module settings. They generally cannot make use of other module settings.
+- [Initialization function](#initialization-function): A function that runs once when the Apostrophe app first starts up.
+- [Customization functions](#customization-functions): Settings via functions that at least have access to the module itself as an argument. They often make use of other settings.
 
 ## Configuration settings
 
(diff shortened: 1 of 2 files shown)


docs/reference/module-api/module-overview.md
```diff
@@ -321,27 +332,59 @@ Each of these function sections takes the module, as `self`, as an argument. Thi
...
+        // Returning the generated piece, exactly as the original `generate`
+        // method does.
+        return piece;
+      }
+    };
+  }
+};
+```
+
+Or this, if extending an [async component function](#components-self):
```
localghost443, line 366: "Or this" feels a bit clunky. Maybe something like 'This example shows'. I don't feel I have an adequate understanding to fully suggest

docs/reference/module-api/module-overview.md
```diff
@@ -386,26 +429,6 @@ Methods included should take a `_super` argument, followed by the normal argumen
...
-        return piece;
-      }
-    };
-  }
-};
-```
-
 ### `components(self)`
 
 This function section returns an object containing functions that power asynchronous template components. These template components allow for asynchronous data requests in normally synchronous template rendering.
```
localghost443, line 434:
```suggestion
Returns an object containing functions that power asynchronous template components. These template components allow for asynchronous data requests in normally synchronous template rendering.
```

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+    };
+  }
+};
+```
+
 #### `extendQueries(self)`
 
+Extend the behavior of existing event handlers (set in `queries`) in the `extendQueries` section. This function must return an object as described in `queries`.
+
+Each extended query builder or method should accept the original function as `_super` followed by its original arguments, if there are any. Extended query builders and methods will be matched with the base class builder or method using the same name. Methods should return data in a similar format to the existing API route.
```
localghost443, line 860:
```suggestion
Each extended query builder or method should accept the original function as `_super` followed by its original arguments. Extended query builders and methods will be matched with the base class builder or method using the same name. Methods should return data in a similar format to the existing API route.
```

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+
 ### `middleware(self)`
 
+Add standard Express middleware to be called on *every* request. The `middleware` function takes the module as an argument and must return an object of [middleware functions](https://expressjs.com/en/guide/using-middleware.html). This is a good place to import third-party middleware into Apostrophe.
```
localghost443, line 864:
```suggestion
Add standard Express middleware to be called on *every* request. The `middleware` function takes the module as an argument and must return an object of [middleware functions](https://expressjs.com/en/guide/using-middleware.html). This is a good place to import third-party middleware.
```

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+
+Note that it can often be simpler to add an event handler or `await` a method in an API route instead.
```
localghost443, line 866:
```suggestion
Note that it is often better to add an event handler or `await` a method in an API route instead.
```
Why?

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+
+`ours` could also be an object with a `before` property and a `middleware` property, in which case it would run `before` the middleware of the module with the specified name
```
localghost443, line 868: Not a suggestion but, `ours`?

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+      }
+    };
+  }
+};
+```
+
 ### `tasks(self)`
 
-## "Cascading" settings
+`tasks` takes the module as an argument and returns an object of functions that define command-line tasks. Task properties include:
```
localghost443, line 918:
```suggestion
`tasks` takes the module as an argument and returns an object of functions that define command line tasks. Task properties include:
```

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+
+
+| Property | Description |
+| ------- | ------- |
+| `usage` | A string describing the task and how to use it. It is printed on the command-line. |
```
localghost443, line 923:
```suggestion
| `usage` | A string describing the task and how to use it. It is printed on the command line. |
```

Side note, I didn't realize that **command prompt** was Windows and **command line** was Unix

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+| `task` | The task function. It maybe asynchronous. |
```
localghost443, line 924:
```suggestion
| `task` | The task function. Can be asynchronous. |
```
or 

```suggestion
| `task` | The task function. It may be asynchronous. |
```

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+| `afterModuleInit` | Set to `true` to run the task after modules are initiated but *before* they are fully "awake".
```
localghost443, line 925: We should not use "awake" for this.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+| `exitAfter` | Only relevant if `afterModuleInit` is `true`. Set to `false` to *avoid* exiting the Apostrophe process on completion. Uncommon. |
+
+Task functions takes the object `argv` as an argument, which includes the arguments passed after the task command. As documented by the [Boring](https://www.npmjs.com/package/boring) utility:
```
localghost443, line 928: Should we link to GitHub instead?

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+
+> When someone types:
```
localghost443, line 930:
```suggestion
> Input:
```

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+>
+> ```
+> node app jump sideways --foo --bar=whee --super-cool=totally
+> ```
+>
+> You get:
```
localghost443, line 936:
```suggestion
> Response:
```
or 

```suggestion
> Output:
```

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+> }
+> ```
+
+```javascript
+// modules/product/index.js
+module.exports = {
+  // ...
+  tasks(self, options) {
+    return {
+      // Since the module is named product, then you can run this CLI task by
```
localghost443, line 953:
```suggestion
      // Since the module is named product, then you can run this command line task by
```
Just trying to be consistent where it makes sense. If we use CLI more, this is fine.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +826,146 @@ module.exports = {
...
+      // typing: `node app product:list` in the CLI.
```
localghost443, line 954:
```suggestion
      // typing: `node app product:list` in the CLI.
```

Same as above

docs/reference/module-api/module-overview.md
```diff
@@ -1,6 +1,9 @@
 # Module properties
 
-Module configuration objects may use the following configuration properties.
+Module configuration objects may use the following configuration properties. The overall categories are broadly defined:
+- [Configuration settings](#configuration-settings): Static module settings. They generally cannot make use of other module settings.
+- [Initialization function](#initialization-function): A function that runs once when the Apostrophe app first starts up.
+- [Customization functions](#customization-functions): Settings via functions that at least have access to the module itself as an argument. They often make use of other settings.
```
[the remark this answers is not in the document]
abea: THERE'S NO CRYING IN BASEBALL

docs/reference/module-api/module-overview.md
```diff
@@ -321,27 +332,59 @@ Each of these function sections takes the module, as `self`, as an argument. Thi
 | [`extendHandlers`](#extendhandlers-self) | Extend base class server-side event handlers |
 | [`queries`](#queries-self-query) | Add database query methods |
 | [`extendQueries`](#extendqueries-self) | Extend base class database query methods |
-| [`middleware`](#middleware-self) | REPLACE-ME |
-| [`tasks`](#tasks-self) | REPLACE-ME |
+| [`middleware`](#middleware-self) | Add standard Express middleware to be called on *every* request. |
+| [`tasks`](#tasks-self) | Add task functions that can be run from the command-line. |
```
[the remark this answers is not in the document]
abea: 👍

docs/reference/module-api/module-overview.md
```diff
@@ -14,6 +17,14 @@ Module configuration objects may use the following configuration properties.
 | [`filters`](#filters) | Object | Yes | Configure piece type filters | Piece |
 | [`columns`](#columns) | Object | Yes | Configure piece type manager columns | Piece |
 
+
+### "Cascading" settings
+
+Many Apostrophe module sections are structured as objects with `add`, `remove` and (sometimes), `group` properties. This pattern allows these settings to "cascade" from the base classes through to project level classes. The subsections help Apostrophe manage inherited options without requiring those inherited options to be re-declared by developers.
+
+If new setting options are needed, they go in `add`. If base class setting options are no longer wanted, they go in `remove`. And in some cases where grouping happens in the interface, they can be organized in `group`.
```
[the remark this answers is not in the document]
abea: This is a pretty limited group and there's some parallelism in the sentence structure. I'd like to keep this one out of a table.

docs/reference/module-api/module-overview.md
```diff
@@ -321,27 +332,59 @@ Each of these function sections takes the module, as `self`, as an argument. Thi
...
-| [`middleware`](#middleware-self) | REPLACE-ME |
-| [`tasks`](#tasks-self) | REPLACE-ME |
+| [`middleware`](#middleware-self) | Add standard Express middleware to be called on *every* request. |
+| [`tasks`](#tasks-self) | Add task functions that can be run from the command-line. |
 
 ### The extension pattern
 
 Several of these sections use an extention pattern (the sections prefixed with "extend"). These sections are used to add functionality to matching features of the module's base class. Functions in `extendMethods` enhance identically named functions in the base class' `methods` section, `extendApiRoutes` enhances API routes from the base class' `apiRoutes`, and so on.
 
 Each individual function included in the extension section takes a `_super` argument in addition to the same arguments as the original function. `_super` is the original function, which should be called within the new extension.
```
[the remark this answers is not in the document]
abea: There are a lot of functions involved. I'll take another swipe at this.

docs/reference/module-api/module-overview.md
```diff
@@ -321,27 +332,59 @@ Each of these function sections takes the module, as `self`, as an argument. Thi
...
+  // ...
   extendMethods(self) {
     return {
-      insert(_super, req, piece, options) {
-        piece.title = `🆕 ${piece.title}`;
+      generate(_super, index) {
+        // Using _super with the original argument to generate a sample piece.
+        const piece = _super(index);
+        // Adding additional functionality.
+        piece.price = Math.random() * 100;
```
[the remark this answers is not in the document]
abea: piece is product.

docs/reference/module-api/module-overview.md
abea: checkIp

docs/reference/module-api/module-overview.md
abea: Nope. This is the published and public version.

docs/reference/module-api/module-overview.md
abea: It's weird, but I'm quoting directly.

docs/reference/module-api/module-overview.md
abea: This is a direct quote. We can change the Boring README if we want it better.

abea:
Good edits, @localghost443

docs/reference/module-api/module-overview.md
```diff
@@ -1,6 +1,9 @@
 # Module properties
 
-Module configuration objects may use the following configuration properties.
+Module configuration objects may use the following configuration properties. The overall categories are broadly defined:
+- [Configuration settings](#configuration-settings): Static module settings. These generally cannot use other module settings.
```
localghost443, line 4: I'm still confused

docs/reference/module-api/module-overview.md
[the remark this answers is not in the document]
localghost443:
```suggestion
Use `add` to add additional settings, `remove` to remove existing settings, and `group` to  organize settings in the editing interface.
```

**or**

```suggestion
Use `add` to add additional settings, `remove` to remove existing ones, and `group` to  organize them in the editing interface.
```

**or** 

```suggestion
- New setting options 👉 `add`.
- Remove setting options 👉 `remove`. 
- To group settings in the editing interface 👉 `group` 
```

**or**

```suggestion
- New setting options → `add`.
- Remove setting options → `remove`. 
- To group settings in the editing interface → `group` 
```

Two list suggestions and two sentence suggestion that shortens it quite a bit and feels a bit less intense (to me)

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +830,146 @@ module.exports = {
...
+          return res.status(403).send('forbidden');
+        }
+        return next();
+      }
+    };
+  }
+};
+```
+
+If the middleware function must run before another named middleware function, set the returned object key to an object with `before` and `middleware` properties. `before` would be set to the name of a module whose middleware must run after the new function, which is now the value of `middleware`.
```
localghost443, line 894: Wondering if we should rephrase this. To me, it sounds like we're saying "You can declare the order by using `before`" or something, but we're phrasing it like "If you want one thing to run before another thing", which seems like a more complicated way to say it

docs/reference/module-api/module-overview.md
localghost443: @abea Fair enough. We probably should, can we do that here and circle back as a small clean up task? Or otherwise leave a note to clean it up in 2 places

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +830,146 @@ module.exports = {
...
+> }
+> ```
+
+```javascript
+// modules/product/index.js
+module.exports = {
+  // ...
+  tasks(self, options) {
+    return {
+      // Since the module is named product, then you can run this command
```
localghost443, line 957:
```suggestion
      // Because the module is named product, you can run this command
```

docs/reference/module-api/module-overview.md
localghost443, line 4: I still trip over this bit:

are we saying that it can't access settings in other modules? or other settings that exist for other modules.

docs/reference/module-api/module-overview.md
abea: > If you want one thing to run before another thing

This is what I'm trying to say, I think.

docs/reference/module-api/module-overview.md
```diff
@@ -14,6 +17,13 @@ Module configuration objects may use the following configuration properties.
 | [`filters`](#filters) | Object | Yes | Configure piece type filters | Piece |
 | [`columns`](#columns) | Object | Yes | Configure piece type manager columns | Piece |
 
+
+### "Cascading" settings
+
+Many Apostrophe module sections are structured as objects with `add`, `remove`, and sometimes `group` properties. This pattern allows these settings to "cascade" from the base classes through to project level classes. The subsections help Apostrophe manage inherited settings without requiring those inherited settings be re-declared by developers.
```
localghost443, line 23:
```suggestion
Many Apostrophe module sections are structured as objects with `add`, `remove`, and sometimes `group` properties. This pattern allows these settings to "cascade" from the base classes to project level classes without requiring those settings to be redeclared.
```

Very nitpicky, sorry last one I swear.

Related question, "Many Apostrophe module sections". We don't say which ones **do** or **dont** have these sections. Is that at all relevant?

docs/reference/module-api/module-overview.md
```diff
@@ -1,6 +1,9 @@
 # Module properties
 
-Module configuration objects may use the following configuration properties.
+Module configuration objects may use the following configuration properties. The overall categories are broadly defined:
+- [Configuration settings](#configuration-settings): Static module settings. Once the module is initialized these settings are fixed and can't access the module itself or any other module's settings.
```
localghost443, line 4:
```suggestion
- [Configuration settings](#configuration-settings): Static module settings. Once the module is initialized, these settings become fixed, and can't access the module itself or any other module's settings.
```

 I had this in my suggestion but it was illegally done.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +829,146 @@ module.exports = {
...
+            console.log(piece.title);
+          }
+        }
+      }
+    }
+  }
+};
+```
 
-What does cascading mean?
```
localghost443, line 919: now I know

docs/reference/module-api/module-overview.md
```diff
@@ -321,27 +331,63 @@ Each of these function sections takes the module, as `self`, as an argument. Thi
...
 
 ### The extension pattern
 
-Several of these sections use an extention pattern (the sections prefixed with "extend"). These sections are used to add functionality to matching features of the module's base class. Functions in `extendMethods` enhance identically named functions in the base class' `methods` section, `extendApiRoutes` enhances API routes from the base class' `apiRoutes`, and so on.
+Several of these sections use an extension pattern. The sections prefixed with `extend` add functionality to base class sections without the prefix. For example:
+
+- functions in `extendMethods` enhance identically named functions in the base class' `methods` section, and
+- `extendApiRoutes` enhances API routes from the base class' `apiRoutes`
 
-Each individual function included in the extension section takes a `_super` argument in addition to the same arguments as the original function. `_super` is the original function, which should be called within the new extension.
```
boutell, line 331: If you don't need to call the original function at all, just redeclare it in `methods` to override it.

docs/reference/module-api/module-overview.md
```diff
@@ -817,27 +759,6 @@ Extend the behavior of existing event handlers (set in `handlers`) in the `exten
...
-        }
-      }
-    }
-  }
-};
-```
-
 ### `queries(self, query)`
 
 The `queries` function is part of the Apostrophe API for querying the database. It takes two arguments: the module (`self`) and the query that will be called (`query`). It must return an object. That object can have two properties:
```
boutell, line 764: This sounds very strongly like something you call (it is "part of the Apostrophe API for querying the database"). That is going to be confusing.

I think it should say:

The `queries` section is used to extend the features of Apotrophe's API for querying the database.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +829,146 @@ module.exports = {
...
+    };
+  }
+};
+```
+
 #### `extendQueries(self)`
 
+Extend the behavior of existing event handlers (set in `queries`) in the `extendQueries` section. This function must return an object as described in `queries`.
+
+Each extended query builder or method should accept the original function as `_super` followed by its original arguments. Extended query builders and methods will be matched with the base class builder or method using the same name. Methods should return data in a similar format to the existing API route.
```
boutell, line 863: It should be mentioend that they need to be nested in `builders` and `methods` sub-sections again.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +829,146 @@ module.exports = {
...
+
 ### `middleware(self)`
 
+Add standard Express middleware to be called on *every* request. The `middleware` function takes the module as an argument and must return an object of [middleware functions](https://expressjs.com/en/guide/using-middleware.html). This is a good place to import third-party middleware.
```
boutell, line 867: ... If it should be called on every request.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +829,146 @@ module.exports = {
...
+
+Note that it is often better to add an event handler or `await` a method in an API route instead.
```
boutell, line 869: If you are thinking about writing your own middleware, it is often better to add an event handler or `await` a method in the appropriate API routes instead.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +829,146 @@ module.exports = {
...
+
+`checkIp` could also be an object with a `before` property and a `middleware` property, in which case it would run `before` the middleware of the module with the specified name
```
boutell, line 871: Should end with a period and probably be formatted as a note.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +829,146 @@ module.exports = {
...
+          return res.status(403).send('forbidden');
+        }
+        return next();
+      }
+    };
+  }
+};
+```
+
+If the middleware function must run before another specific module's middleware, set the returned object key to an object with `before` and `middleware` properties. `before` would be set to the name of the module whose middleware must run after the new function, which is now the value of `middleware`.
```
boutell, line 893: I think this is said twice.

boutell: This treatment is the better one.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +829,146 @@ module.exports = {
...
+      }
+    };
+  }
+};
+```
+
 ### `tasks(self)`
 
-## "Cascading" settings
+`tasks` takes the module as an argument and returns an object of functions that define command line tasks. Task properties include:
```
boutell, line 921: I guess you don't want to say "object of methods" because that is easily confused with apostrophe module methods. OK.

docs/reference/module-api/module-overview.md
```diff
@@ -908,12 +829,146 @@ module.exports = {
...
+| `task` | The task function. Can be asynchronous. |
+| `afterModuleInit` | Set to `true` to run the task after modules are initiated but *before* they are fully active.
+| `exitAfter` | Only relevant if `afterModuleInit` is `true`. Set to `false` to *avoid* exiting the Apostrophe process on completion. Uncommon. |
+
+Task functions takes the object `argv` as an argument, which includes the arguments passed after the task command. As documented by the [Boring](https://www.npmjs.com/package/boring) utility:
+
+> Input:
+>
+> ```
+> node app jump sideways --foo --bar=whee --super-cool=totally
```
boutell, line 936: The task name will be in `_` as the first argument, and the example should be formatted like a plausible apostrophe task.

boutell (changes requested):

docs/reference/module-api/module-overview.md
localghost443:
```suggestion
If the middleware function must run before another specific module's middleware, set the returned object key to an object with `before` and `middleware` properties. `before` would be set to the name of the module whose middleware must run after the new function. The new function will become the value of `middleware`.
```

docs/reference/module-api/module-overview.md
localghost443:
```suggestion
Note that when authoring your own middleware, it is often better to add an event handler or `await` a method in an API route instead.
```

docs/reference/module-api/module-overview.md
abea: This one doesn't seem right to me. They don't become fixed after initialization. After initialization they are fixed.

docs/reference/module-api/module-overview.md
abea: addressed in #47

docs/reference/module-api/module-overview.md
abea: Definitely adding the period. This is leading directly into the next thing and introduces a new feature of this section, so I think it should be a normal paragraph.

docs/reference/module-api/module-overview.md
abea: some of this is me copying other text lazily.

docs/reference/module-api/module-overview.md
abea: Ben bumped on this as well. I was trying to avoid duplication by quoting Boring, but it's confusing.

localghost443 (approved):

docs/reference/module-api/module-overview.md
```diff
@@ -933,17 +930,17 @@ Task functions takes the object `argv` as an argument, which includes the argume
 > Input:
 >
 > ```
-> node app jump sideways --foo --bar=whee --super-cool=totally
+> node app my-module:runIt taskOption --foo --bar=baz --use-color=green
```
boutell, line 933: Apostrophe tasks are kebab-case. Sorry for the nitpick.

docs/reference/module-api/module-overview.md
```diff
@@ -933,17 +930,17 @@ Task functions takes the object `argv` as an argument, which includes the argume
...
 > ```
 >
 > Response:
 >
 > ```
 > {
->   _: [ "jump", "sideways"],
+>   _: [ "runIt", "taskOption"],
```
boutell, line 940: It'll be `my-module:runIt`, not just `runIt`. It's the actual command line arguments from `boring` (everything after app.js), the task module doesn't try to get cute with it in any way.

boutell (changes requested):

docs/reference/module-api/module-overview.md
abea: This is addressed in the `::: warning` note at the end of the section.

docs/reference/module-api/module-overview.md
abea: 👍

boutell (approved):

PR #52, opened 2021-03-25
PRO-994 Intro module

abea:
Adds an intro section on modules


docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
 
 ## The module system
 
-Vestibulum id ligula porta felis euismod semper. Maecenas faucibus mollis interdum. Cras mattis consectetur purus sit amet fermentum. Fusce dapibus, tellus ac cursus commodo, tortor mauris condimentum nibh, ut fermentum massa justo sit amet risus. Donec ullamcorper nulla non metus auctor fringilla. Cras mattis consectetur purus sit amet fermentum. Donec ullamcorper nulla non metus auctor fringilla.
+As you learn to use ApostropheCMS, you'll quickly see how all server-side code is organized in a system of "modules." Modules are the building blocks we use to build Apostrophe projects — as well as the CMS itself.
```
localghost443, line 19:
```suggestion
In Apostrophe, all server-side code is organized in a system of "modules." Modules are the building blocks we use to build Apostrophe projects — as well as the CMS itself.
```
Removed the first bit because it's not necessary to understanding, imo. I also wonder if there is a more direct way to say 'server-side code'....is it implicit? Idk. 

```suggestion
Apostrophe is organized in a system of modules. Modules are the building blocks we use to build Apostrophe projects — as well as the CMS itself.
```

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+Each module defines a specific set of functionality, from governing the database connection to setting the fields in a blog post. Common Apostrophe project modules will define custom content types, page types, or editable widget types.
```
localghost443, line 21:
```suggestion
Common Apostrophe project modules will define custom content types, page types, or editable widget types.
```
I removed the first sentence because
1. It seems implicite to me that module define functionality,
2. "governing the database connection" feels abstract and a bit scary.
3. The big value pay off is the second sentence.

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+All modules use [the same API](/reference/module-api/). That shared foundation means that you can give your blog post module powerful features, such as custom command line tasks and API routes. It's the same foundation the core team uses to build the CMS, so it is well tested and designed to be as intuitive as possible.
+
+### Setting up a module
+
+Modules are organized in a folder at the root of a project appropriately named `modules`. Each module gets its own directory inside that with an `index.js` file that contains its configurations. So the blog post code would being in the file:
```
localghost443, line 27:
```suggestion
Modules are organized in a folder at the root of a project appropriately named `modules`. Each module has a dedicated directory with an `index.js` file that contains its configurations. So a blog post module would be in the file:
```
1. We say they're organized "in a folder", and then say "gets its own directory inside that", which feels redundant.
2. Typo

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+```
+modules/blog-post/index.js
+```
+
+The module code itself would then be defined in an object, assigned to `module.exports`. This may look familiar if you know [CommonJS](https://nodejs.org/api/modules.html#modules_modules_commonjs_modules) patterns.
```
localghost443, line 33: In this section, I'm not sure what "the module code itself" means. We haven't really defined it. Earlier, we said the index js is for "its configurations". I think we should repeat that here, and give a really small example of a basic config within the object.

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+```js
+// app.js
+require('apostrophe')({
+  modules: {
+    'blog-post': {}
+  }
+});
+```
+
+There is more to do to make modules useful, but that is all it takes to get them running.
```
localghost443, line 53:
```suggestion
```
I think we could remove this guy or otherwise point them in a direction.

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+### Module inheritance
+
+Inheritance is the glue of the module system. Every module, other than the root module, extends another one. This means that your blog post module, which extends the "piece type" module, has a huge set of features you never have to write.
```
localghost443, line 57: Link to piece type module when we have a chance. Or this might be a future opportunity for a tooltip to go over "piece type"

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+```javascript
+// modules/blog-post/index.js
+module.exports = {
+  extend: '@apostrophecms/piece-type'
+};
+```
+
+More than that, it means you can easily customize that core ["piece type"](/reference/glossary.md#piece) module (`@apostrophecms/piece-type`) in your project. That will not only apply changes to the project's blog post content type, but also every piece type in Apostrophe core.
```
localghost443, line 66: I see you linked piece type here, but since the last paragraph was the first use, I would do it there instead (or in addition to).

```suggestion
More than that, it means you can easily customize that core ["piece type"](/reference/glossary.md#piece) module (`@apostrophecms/piece-type`) in your project. Editing the core piece type will apply changes to every piece type  in Apostrophe core as well as piece types you've created yourself..
```

My attempt at removing "that will not only". will + not + only feels a bit confusing. "That" could be better defined, as "If you edit" or "Editing the"

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+Configuring a core or installed module is as simple as creating an `index.js` file for the module in your project. For example, we might want to log the title of every piece when it is published. We would then create the file:
```
localghost443, line 68: Is this a common use case?

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+        logPublished (req, data) {
+          console.log(`Published ${data.published.title}`);
+        }
+      }
+    };
+  }
+};
+```
+
+Since every piece type extends that module, they all get the benefits of changes to it. And in the project code we only need to include our changes. All the rest of the module's code is untouched.
```
localghost443, line 91: We technically just said this, but it is kind of nice to end with.

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+The rest of the documentation will include many specific, practical examples of these ideas. For now a couple of main takeaways are:
```
localghost443, line 93:
```suggestion
The rest of the documentation will include many specific, practical examples of these ideas. For now the main takeaways are:
```
couple means "two". There are 3, but really, these are all of them, imo.

localghost443 (changes requested):

docs/guide/introduction.md
abea: The point isn't that they defined functionality, but that each one governs a specific set of functionality (they don't share responsibility for things).

abea: I'll reverse order to not lead with the complicated one. My point with this is to point out that they really do manage everything, from the simple to complex.

docs/guide/introduction.md
abea: It's not, but it's a simple one. Real use cases would introduce additional features that I didn't want to bring in so as to focus on the point here.

docs/guide/introduction.md
abea: It means two in some contexts, but generally means an indefinite small number.

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+The module configuration is in an object, assigned to `module.exports`. This may look familiar if you know [CommonJS](https://nodejs.org/api/modules.html#modules_modules_commonjs_modules) patterns.
+
+```javascript
+// modules/blog-post/index.js
+module.exports = {
+  // ...
+}
+```
+
+The final step to this minimalist module example is to tell Apostrophe that it should be turned on, or instantiated. That is done in the main application file, `app.js` in its own `modules` property.
```
localghost443, line 42:
```suggestion
The final step is to tell Apostrophe that it should be turned on, or instantiated. That is done in the main application file, `app.js` in its `modules` property.
```

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+    'blog-post': {}
+  }
+});
+```
+
+The module API supports many different configuration options. See the [module API reference](/reference/module-api/module-overview.md) for more detail.
+
+### Module inheritance
+
+Inheritance is the glue of the module system. Every module, other than the root module, extends another one. This means that your blog post module, which extends the ["piece type"](/reference/glossary.md#piece) module, has a huge set of features you never have to write.
```
localghost443, line 57:
```suggestion
Inheritance is the glue of the module system. Every module, other than the root module, extends another module. This means that your blog post module, which extends the ["piece type"](/reference/glossary.md#piece) module, comes with a huge set of features you never have to write.
```

localghost443 (changes requested):

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
 
 ## The module system
 
-Vestibulum id ligula porta felis euismod semper. Maecenas faucibus mollis interdum. Cras mattis consectetur purus sit amet fermentum. Fusce dapibus, tellus ac cursus commodo, tortor mauris condimentum nibh, ut fermentum massa justo sit amet risus. Donec ullamcorper nulla non metus auctor fringilla. Cras mattis consectetur purus sit amet fermentum. Donec ullamcorper nulla non metus auctor fringilla.
+Apostrophe code is organized in a system of modules. Modules are the building blocks we use to build Apostrophe projects — as well as the CMS itself.
```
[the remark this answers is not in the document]
abea: I don't know about this.

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+Each module defines a specific set of functionality, from configuring blog post fields  to governing internationalization. Common Apostrophe project modules will define custom content types, page types, or editable widget types.
```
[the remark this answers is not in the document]
abea: I'm going for a range of complexity. Alternate examples are welcome.

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+
+All modules use [the same API](/reference/module-api/). That shared foundation means that you can give your blog post module powerful features, such as custom command line tasks and API routes. It's the same foundation the core team uses to build the CMS, so it is well tested and designed to be as intuitive as possible.
+
+### Setting up a module
+
+Modules are organized in a folder at the root of a project appropriately named `modules`. Each module has a dedicated directory inside that with an `index.js` file that contains its configurations. So you would define a blog post module in the file:
```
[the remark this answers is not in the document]
abea: I'm trying not to be too dry, hence the "appropriately."

docs/guide/introduction.md
```diff
@@ -16,7 +16,84 @@ In short, it recognizes pain points on both sides of the CMS experience and reli
...
+Apostrophe code is organized in a system of modules. Modules are the building blocks we use to build Apostrophe projects — as well as the CMS itself.
+
+Each module defines a specific set of functionality, from configuring blog post fields to governing internationalization. Common Apostrophe project modules will define custom content types, page types, or editable widget types.
+
+All modules use [the same API](/reference/module-api/). That shared foundation means that all have access to powerful features, such as custom command line tasks and API routes. It's the same API the core team uses to build the CMS, so it is well tested and designed to be as intuitive as possible.
+
+### Setting up a module
+
+Modules are organized in a folder at the root of a project appropriately named `modules`. Each module has a dedicated directory with an `index.js` file that contains its configurations. So you would define a blog post module in the file:
```
localghost443, line 27:
```suggestion
Modules are organized in a folder at the root of a project named `modules`. Each module has a dedicated directory with an `index.js` file that contains its configurations. So you would define a blog post module in the file:
```

localghost443 (changes requested):

localghost443 (approved):

PR #55, opened 2021-03-26
PRO-937 Pieces basics page

abea:
In addition to adding a basics page for pieces, I've rearranged the guide pages. These are the alpha-stage pages, but they represent the order we'll go for and I think they still work as they are for now.

When I add the pieces pages section it may be as a subsection of this intro page rather than a separate page. The piece basics isn't too long and the two topics are extremely connected.


docs/guide/pieces.md
```diff
@@ -1,396 +1,85 @@
----
-title: "Pieces"
----
-
 # Pieces
-What is a Piece?
 
-As experienced A2 developers know, a piece type module defines structure and behavior for a single type of content, such as blog posts, products, or events. If you have hundreds of documents of the same type, that's a piece. If you want to let the user view them in a certain sorting order, that's a piece. If the same item might be appropriate to display on many different pages, that's a piece. Blog posts, events and products are all excellent examples. Less obvious examples include "category" or "tag" content types that serve to create a taxonomy for other pieces.
+"Piece types" are classifications for standalone, structured content. Each database document created with a piece type module is known as a "piece". The piece type defines how pieces are structured, how end users interact with them, and how they might interact with other content. Blog posts, events, and people profiles are all common examples of piece types.
```
localghost443, line 3:
```suggestion
"Piece types" are classifications for structured content. Events, products, and author profiles are all common examples of piece types. Individual records of a piece type are known as "pieces".
```

**Each database document created with a piece type module is known as a "piece".**  -- this feels like it could have value, but not here. We don't need to talk directly about the database, this is implied. It might be beneficial to have a section on 'how things are stored' in the db, for developers who wish to know.

**The piece type defines how pieces are structured, how end users interact with them, and how they might interact with other content** -- I think this is implied, so I suggest removal. We know that a piece is 'structured content', so that part isn't necessary. A CMS is about creating content that end users can interact with, so I think thats also implied. 

Removed 'blog posts' from this because its just a bit or difficult to understand. Later, I think we use the example of "articles in a blog" to define 'pieces', which is more clear.

docs/guide/pieces.md
```diff
@@ -1,396 +1,85 @@
...
+
+Use cases for pieces include:
+- Content that will be listed using pagination, such as **articles in a blog**
```
localghost443, line 6:
```suggestion
- Articles in a blog
```
Pagination isn't important here, imo.

docs/guide/pieces.md
```diff
@@ -1,396 +1,85 @@
...
+- Content of a particular type that will help to organize other content, such as **teams that connect to people pieces**
+- Content that will be selectively displayed based on end user input, such as **products on a store's website**
```
localghost443, line 8:
```suggestion
- Individual products in a store
```
The selectivity doesn't seem critical to this example? We're always on a website in these docs.

docs/guide/pieces.md
```diff
@@ -1,396 +1,85 @@
...
 # Pieces
-What is a Piece?
 
-As experienced A2 developers know, a piece type module defines structure and behavior for a single type of content, such as blog posts, products, or events. If you have hundreds of documents of the same type, that's a piece. If you want to let the user view them in a certain sorting order, that's a piece. If the same item might be appropriate to display on many different pages, that's a piece. Blog posts, events and products are all excellent examples. Less obvious examples include "category" or "tag" content types that serve to create a taxonomy for other pieces.
+"Piece types" are classifications for standalone, structured content. Each database document created with a piece type module is known as a "piece". The piece type defines how pieces are structured, how end users interact with them, and how they might interact with other content. Blog posts, events, and people profiles are all common examples of piece types.
+
+Use cases for pieces include:
+- Content that will be listed using pagination, such as **articles in a blog**
+- Content of a particular type that will help to organize other content, such as **teams that connect to people pieces**
```
localghost443, line 7: This one is confusing. Can we use something like "The **author** of an article".

docs/guide/pieces.md
```diff
@@ -1,396 +1,85 @@
...
+- Content that will be selectively displayed based on end user input, such as **products on a store's website**
 
-Apostrophe provides both pieces and "[piece pages](piece-pages.md)," which afford your users a way to browse, filter, paginate and discover pieces of a particular kind. For more information about using piece pages, see the [piece pages section](piece-pages.md).
+Pieces are distinct from "pages" in Apostrophe in that they do not inherently have an internal hierarchy: generally speaking, one piece is not the "parent" of another of the same type. Pieces also do not necessarily have dedicated web pages for each one. In the "teams" and "people" example, there may be a web page for each team, but the people are only displayed within those team pages -- not on their own pages.
```
localghost443, line 10:
```suggestion
Pieces do not always have a dedicated slug or web page. For example, in the "author" and "article" example, there will likely be a page for every article, but not a page for every author. 
```

I dont really understand what 'internal hierarchy' means here, or if its necessary to call attention to the fact that they are orphan documents. To me, this implies that if you deleted a piece-type, everything would still be fine, but thats not the case.

localghost443 (changes requested):

docs/guide/pieces.md
abea: I don't think we should consider so much to be obvious to a new developer. In some systems the content types only really define the fields available. In Apostrophe we make much more than that available to configure, so I suggest the more complete picture.

docs/guide/pieces.md
abea: It refers to the page tree and pieces not having inherent relationships with one another the way pages do.

localghost443 (approved):

PR #56, opened 2021-03-31
PRO-936 Adds page essentials page

abea:


docs/guide/pages.md
```diff
@@ -0,0 +1,239 @@
+# Pages and page types
+
+Every page on the an Apostrophe website is assigned a **"page type"**. The page type tells Apostrophe **what template to use** to render the page and **what configuration to apply**. Configurations will often at least include the content schema for the page type.
+
+## Creating a page type
+
+Apostrophe core only includes a "Home page" type with minimal functionality. You will likely need your own page types, which you create by adding modules that extend `@apostrophecms/page-type` and instantiating them in `app.js`.
+
+```js
+// modules/default-page/index.js
```
localghost443, line 10: Is default page like, really a thing, or is it just a habit we have? Should we use a better example?

docs/guide/pages.md
```diff
@@ -0,0 +1,239 @@
...
+### We are inserting page template markup in a template block
+
+```django
+{% block main %}
+{% endblock %}
+```
+
+Apostrophe uses the Nunjucks template language, which has a [block system](https://mozilla.github.io/nunjucks/templating.html#block) for injecting markup into lower-level templates. The block system involves placing a `block` tag in the root-level template file, then using those blocks in higher-level templates to insert markup.
+
+In the Default page example, the layout template's `beforeMain` and `afterMain` blocks are not replaced, so rendered Default pages use that markup as-is. The Default page template does use the `main` block, so the layout template's version is replaced.
```
localghost443, line 125:
```suggestion
```
I almost dont think we need this. I felt like the example did a fairly good job of illustrating everything, but when I got here I found myself tripping over this paragraph, and in the end realizing it said what I already perceived to be true.

localghost443 (changes requested):

docs/guide/pages.md
```diff
@@ -0,0 +1,239 @@
...
+};
+```
+
+See the [content schema](/guide/introduction.md#content-schemas) page for more on configuring fields.
+
+## Page template essentials
+
+You will need to add a template file for each page type. The only exception to that rule is if your page type module extends another page type that already has a template.
+
+Page templates are added in a `views` directory in the module as `page.html`. The template for the previous example's default page would be `modules/default-page/views/page.html`. A very simple page template for the Default page might look like this:
```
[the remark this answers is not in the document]
abea: "for"

docs/guide/pages.md
abea: "Since we used the main block here it will overwrite the layout template main block."

localghost443 (approved):

PR #58, opened 2021-04-01
PRO-1070 Adds note about setting session secret

abea:


localghost443 (approved):

boutell (approved):

boutell:
Cool. Maybe someday the session secret should be persisted in the database and read at startup, that would be an option to avoid this without replacing values in the CLI.

PR #60, opened 2021-04-01
PRO-932 Widget essentials

abea:
This is one of three pages about widgets. There will be additional pages covering core widgets (RTE, image, video, etc) and another on building new widget types.


docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,147 @@
+# Areas and widgets
+
+**Areas** are core to Apostrophe's in-context editing experience. They are special fields where editors can add one or more content widgets. A **widget** is a section of structured content, such as a block of rich text, an image slideshow, or a grid of featured products. Together, these two features let editors add custom and advanced content to a website, move it around, and edit it &ndash; all within a defined modal that maintains the content design.
```
localghost443, line 3: Very good but got lost at the modal part

docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Basic area configuration
+
+Like other fields, you will configure an area as part of the [field schema](/guide/content-schema.md) for a page or piece type. Apostrophe comes with several widgets that you can use right away, including ones for adding rich text, inserting an image, and embedding a video.
```
localghost443, line 7: I think we could add a sentence about "areas accept widgets". I never advocate for adding sentence but it may make sense here between the first and second.

localghost443: alternatively I see you say this more or less below, so maybe we could nix the second sentence.

docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,147 @@
...
+      mainArea: {
+        label: 'Main page content'
+        fields: ['main']
+      }
+    }
+  }
+};
+```
+
+The configuration above shows a landing page type with one area field named `main`. Every area needs a `widgets` option that identifies the widget types allowed there. This example includes three widget of the core widget types, which would then be available to editors to add and arrange.
```
localghost443, line 39:
```suggestion
The configuration above shows a landing page type with one area field named `main`. Every area requires a `widgets` option that configures the allowed widget types. This example includes three of the core widget types available.
```

docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+![The landing page main area with the menu open, showing available widgets](/images/area-in-context.jpg)
+
+### Leave `-widget` out of area configuration
+
+If you dig into the Apostrophe code, or read the guide to custom widget types, you may already know that widget module names end with `-widget`, such as `@apostrophecms/rich-text-widget`. This suffix is left off when configuring areas. The area knows these are widgets already and it does not seem necessary to require developers don't need to write it over and over.
```
localghost443, line 45: Oooh

docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+```django
+{% area data.page, main %}
+```
+
+The template tag knows to use the area data on `data.page.main`, check for the widgets allowed in that area, and render the area using the correct widget templates.
+
+### Passing context options
+
+Any widget options that the server needs to know about must be included in the area field configuration. For example, you can configure the rich text widget to use particular formatting controls. The server needs to know those to properly validate user input.
```
localghost443, line 97: How does a user know if the server needs to know about it?

docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,147 @@
...
+} %}
+```
+
+The object following `with` should include keys matching widget type names, without the `-widget` suffix (e.g., the `@apostrophecms/image`). The context template will pass those options into the proper widget template as `data.contextOptions`. In the example above, the core image widget template, *and only that template*, would be able to use the data as:
+
+```django
+{{ data.contextOptions.sizes }}
+```
+
+Any context options for widget types not allowed in the area are ignored. So in this next example, since the area is not configured to use a `fake-widget`, that option would not be passed into any widget templates.
```
localghost443, line 139: Not sure if we need an example of fake, but its good to note that they're ignored.

localghost443 (changes requested):

docs/guide/areas-and-widgets/README.md
abea: Should be "model"

docs/guide/areas-and-widgets/README.md
abea: Move "Every area needs a `widgets`" up here?

docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }
+}
+```
+
+::: note
+Learn more about rich text options in [the section on core widgets](/guide/areas-and-widgets/core-widgets.md).
+:::
+
+In other situations, you may need to **pass the widget *template* options that only apply to a specific context** and not everwhere the area is shown. One example of this is the `sizes` attribute for the core image widget's `img` tag. Since that attribute tells browsers which file versions to use in a responsive image, it may be different when the image is a small thumbnail as opposed to when it is a larger featured photo.
```
localghost443, line 119:
```suggestion
In other situations, you may need to **pass the widget *template* options that only apply to a specific context**. One example of this is the `sizes` attribute for the core image widget's `img` tag. Since that attribute tells browsers which file versions to use in a responsive image, it may be different when the image is a small thumbnail as opposed to when it is a larger featured photo.
```

localghost443 (changes requested):

docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,137 @@
+# Areas and widgets
+
+**Areas** are core to Apostrophe's in-context editing experience. They are special fields where editors can add one or more content widgets. A **widget** is a section of structured content, such as a block of rich text, an image slideshow, or a grid of featured products. Together, these two features let editors add custom and advanced content to a website, move it around, and edit it &ndash; all within a model that maintains the content design.
+
+## Basic area configuration
+
+Like other fields, you will configure an area as part of the [field schema](/guide/content-schema.md) for a page or piece type. The following example shows a landing page type with one area field named `main`. Every area requires a `widgets` option to configure the allowed widget types. This example includes three core widget types.
```
localghost443, line 7:
```suggestion

Like other fields, area fields are configured as part of the [field schema](/guide/content-schema.md) for a page or piece type. The following example shows a landing page type with one area field named `main`. Every area requires a `widgets` option to configure the allowed widget types. This example includes three core widget types.
```

docs/guide/areas-and-widgets/README.md
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }
+}
+```
+
+::: note
+Learn more about rich text options in [the section on core widgets](/guide/areas-and-widgets/core-widgets.md).
+:::
+
+In other situations, you may need to **pass the widget *template* options that only apply to a specific context**. One example of this is the `sizes` attribute for the core image widget's `img` tag. Since that attribute tells browsers which file versions to use in a responsive image, it may be different when the image is a small thumbnail as opposed to when it is a larger featured photo.
```
localghost443, line 119: should we link to sizes here?

localghost443 (changes requested):

localghost443 (approved):

PR #62, opened 2021-04-05
PRO-1127 Widgets core

abea:
adds core widgets page

### DIFF
--- a/docs/guide/areas-and-widgets/core-widgets.md
+++ b/docs/guide/areas-and-widgets/core-widgets.md
@@ -39,20 +39,20 @@ widgets: {
 
 ### Configuring the toolbar
 
-To add formatting tools to the rich text toolbar, add their names to the `toolbar` array. The available tools include:
+To add formatting tools to the rich text toolbar, add their names to the `toolbar` array. The available formatting options include:
 
 | Tool name | What is it? |
 | --------- | ----------- |
-| 'styles' | A drop down list of text styles, allowing different HTML tags and CSS classes (see ["Configuring text styles"](#configuring-text-styles) below) |
+| 'styles' | A list of text styles, allowing different HTML tags and CSS classes (see ["Configuring text styles"](#configuring-text-styles) below) |
 | `'bold'` | Bold text |
 | `'italic'` | Italicize text |
-| `'strike'` | Strikethrough |
+| `'strike'` | Strikethrough text |
 | `'link'` | Add a link |
 | `'horizontal_rule'` | Add a visual horizontal rule |
-| `'bullet_list'` | Convert text to a bulleted list |
-| `'ordered_list'` | Convert text to a numbered list |
-| `'blockquote'` | Convert text to a block quote |
-| `'code_block'` | Convert text to a code quote |
+| `'bullet_list'` | Bulleted list |
+| `'ordered_list'` | Numbered list |
+| `'blockquote'` | Block quote |
+| `'code_block'` | Code block |
 | `'undo'` | Undo the last change |
 | `'redo'` | Redo the last undone change |
 | `'|'` | Add a visual separator to the toolbar (not a formatting action) |
@@ -61,7 +61,7 @@ To add formatting tools to the rich text toolbar, add their names to the `toolba
 
 ### Configuring text styles
 
-When you add the `'styles'` formatting tool, you can configure an array of text styles for editors to apply. These must include an HTML tag and a label for the menu. They may also include a CSS class.
+When you add the `'styles'` formatting tool, you can configure an array of text styles. These must include an HTML tag and a label for the menu interface. They may also include a CSS class.
 
 A single style including class might look like:
 
@@ -81,9 +81,7 @@ Including a class with a style will not automatically apply any styles. You stil
 
 <!-- TODO: Link to how-to about configuring sanitize-html for pasting in rich text -->
 
-### Default configuration
-
-There is default configuration for rich text widgets so you do not necessarily need to configure yours. That configuration is:
+### Default rich text configuration
 
 ```javascript
 {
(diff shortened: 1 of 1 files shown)


docs/guide/areas-and-widgets/core-widgets.md
```diff
@@ -1,128 +1,263 @@
...
+| `'code_block'` | Convert text to a code quote |
+| `'undo'` | Undo the last change |
+| `'redo'` | Redo the last undone change |
+| `'|'` | Add a visual separator to the toolbar (not a formatting action) |
+
+<!-- TODO: Add a link to the how-to on adding your own tools when available. -->
+
+### Configuring text styles
+
+When you add the `'styles'` formatting tool, you can configure an array of text styles for editors to apply. These must include an HTML tag and a label for the menu. They may also include a CSS class.
```
localghost443, line 64: should we offer a table here?

docs/guide/areas-and-widgets/core-widgets.md
```diff
@@ -1,128 +1,263 @@
...
+### Customizing responsive image sizes
 
-::: tip Note:
-The image widget now only accepts one image. A3 comes with a still image widget, but because we are [less opinionated on the front end](front-end-assets.md), it doesn't come with a slideshow widget like in A2.
-:::
+<!-- TODO: link to attachment module srcset method when reference is available. -->
+The image widget's default `srcset` attribute for responsive behavior assumes the image is roughly the same width as the viewport. This will help reduce download times even if the display size is smaller, but you can make responsive loading more accurate by providing [the `sizes` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-sizes) value.
 
-### Sizing the image to suit the placement
+This configuration is applied as a template option, rather than in server-side code, since the value generally depends on the context where it's appied. In the template's `area` tag, include a `sizes` [context option](/guide/areas-and-widgets/#passing-context-options) for the image widget. The option value should be the actual HTML attribute value.
```
localghost443, line 158:
```suggestion
This configuration is applied as a template option. In the template's `area` tag, include a `sizes` [context option](/guide/areas-and-widgets/#passing-context-options) for the image widget. The option value should be the actual HTML attribute value.
```

docs/guide/areas-and-widgets/core-widgets.md
```diff
@@ -1,128 +1,263 @@
...
+## Video widget
 
-## Configuring the video widget
+The core video widget accepts a video URL and fetches the embed code to display it. Most major video hosts are supported by default. The widget uses the oEmbed data format, so you can find a full list of supported services [on the oEmbed website](https://oembed.com/#section7).
 
-While the video widget looks better out of the box, you can configure a `className` option for that as well if you wish.
+## HTML widget
 
-## Configuring the HTML widget
+**...or, how to get access to the editing interface when embedded HTML breaks it.**
```
localghost443, line 206:
```suggestion
```

docs/guide/areas-and-widgets/core-widgets.md
```diff
@@ -1,128 +1,263 @@
...
+https://example.net/broken-page?safemode=1
+```
 
-There's nothing to configure! But, note that if you paste a bad embed code that breaks the tag balancing of the markup or otherwise damages the page, you will need a way to get control back.
 
 To do that, access the page with `?safemode=1` at the end of the URL. Then you will be able to edit the widget and remove the offending content.
+
+## Setting a CSS class on core widgets
+
+It is a front end development best practice to have a consistent `class` naming system, regardless of what that system is. There are two options to set classes on core widgets to support this practice and avoid overwriting the whole template to do so.
```
localghost443, line 223:
```suggestion
There are two options to set classes on core widgets.
```
Idk I just think we could cut to the chase here. Leave best practice out with showing rather than telling.

localghost443 (changes requested):

docs/guide/areas-and-widgets/core-widgets.md
```diff
@@ -1,128 +1,263 @@
...
 :::
 
-## Configuring the image widget
+<!-- TODO: Link to how-to about configuring sanitize-html for pasting in rich text -->
 
-The image widget does work right out of the box, but notice that images can push beyond the page. A3 does not impose any front-end opinions regarding widgets, and thus it's necessary to configure them properly with CSS classes for styling. In this example, we'll configure the image-widget to have a class.
+### Default configuration
 
-Start by creating a directory for project-level configuration of the `@apostrophecms/image-widget` module. Using your terminal:
+There is default configuration for rich text widgets so you do not necessarily need to configure yours. That configuration is:
```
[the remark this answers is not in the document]
abea: I'll tone it down, but make clear what we're specifically talking about.

docs/guide/areas-and-widgets/core-widgets.md
```diff
@@ -1,128 +1,263 @@
...
+  ]
+},
 ```
 
-Then, you can configure it by creating an `index.js` file in that directory:
+If your rich text settings only include one of the two sections (`toolbar` and `styles`), the default for the other will be used. In other words, if your configuration only changes `styles`, the `toolbar` option from the defaults will apply.
 
-```js
-// modules/@apostrophecms/image-widget/index.js
+**You can also set your own defaults.** If you use the same rich text options in all or most of your areas, you can configure these option on the `@apostrophecms/rich-text-widget` module as opposed to on individual areas.
```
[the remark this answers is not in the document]
abea: I think this might lose the distinction.

docs/guide/areas-and-widgets/core-widgets.md
abea: Haters gonna hate.

localghost443 (approved):

PR #63, opened 2021-04-07
PRO-1128 Custom widgets page

abea:


docs/guide/areas-and-widgets/custom-widgets.md
```diff
@@ -1,116 +1,206 @@
...
+    'two-column-widget': {}
+  }
+});
 ```
 
+## Widget templates
+
+Before using the new widget type, it needs a template file, `widget.html`, in the module's `views` directory. A simple template for the two column widget might look like:
+
 ```django
```
localghost443, line 71: should we keep Django here? is that still the most accurate?

localghost443 (changes requested):

docs/guide/areas-and-widgets/custom-widgets.md
abea: Yep.

localghost443 (approved):

PR #64, opened 2021-04-08
PRO-1149 conditional fields

abea:


docs/guide/major-changes.md
```diff
@@ -121,7 +121,7 @@ Methods are now declared in the `methods` section. Using the "super pattern" to
 
 ### `components` and `extendComponents`
 
-The new [async components](async-components.md) feature lets you fetch content on the fly from inside your page templates. Async components are a recommended replacement for most common uses of the old `apostrophe-pages:beforeSend` promise event handler. Async components are async functions whose return values are passed to a Nunjucks template of the same name. The result is rendered at the point in the page where `{% component "module-name:componentName" with { data... } %}` was called. The async function receives `(req, data)` so it can work with information passed by the template.
```
abea, line 124: Only link fixes here. No content review please.

docs/guide/advanced/conditional-fields.md
```diff
@@ -0,0 +1,79 @@
...
+    // 👇 Two conditions that both must be met
+    if: {
+      seenMovie: true,
+      relationship: 'none'
+    }
+  }
+}
+```
+
+Additional conditional options will be added in the future.
```
boutell, line 79: We should cover `$or` as it has already been implemented and delivered.

boutell (changes requested):

docs/guide/advanced/conditional-fields.md
abea: Got it. I wasn't sure if it had been.

docs/guide/advanced/conditional-fields.md
```diff
@@ -33,6 +33,8 @@ add: {
 }
 ```
 
+## Multiple required conditions
+
 **The `if` setting may contain more than one condition.** When there is more than one, all conditions must be met before the field will be active.
 
 In the next example, `seenMovie` must be `true` *and* `relationship` must be `'none'` for the rating field to appear.
```
boutell, line 40: For purposes of an example in our documentation, `relationship` is a potentially confusing field name.

boutell (approved):
Approving, but I do have a concern about the field name `relationship` being potentially confusing. If you need to move on it's cool.

docs/guide/advanced/conditional-fields.md
```diff
@@ -0,0 +1,103 @@
+---
+permalink: '/guide/conditional-fields'
+---
+
+# Conditional fields
+
+When defining a field schema, you may make fields conditional on other field values using the `if` setting. Until the `if` setting conditions are met, the field will be hidden. The conditions must be an object with keys *matching the names of other fields in the same schema*. The condition values must match the sibling field values *exactly* to pass.
+
+::: note
+Because strict equivalence is required for conditions, fields used in conditions must have values that are strings, numbers, or booleans. This is subject to change.
```
localghost443, line 10:
```suggestion
Because strict equivalence is required for conditions, fields used in conditions must have values that are strings, numbers, or booleans.
```

docs/guide/advanced/conditional-fields.md
localghost443: I agree, I'm not sure what would go here. It could be confused with a relationship field (unlikely, but still). Also, what kind of relationships do most people have with filmmakers. By most people, I mean people that aren't stars living in the hills of LA, like myself.

localghost443 (changes requested):
I feel the same as tom about the 'relationship' example. I also feel that we should have a table of some sort in here outline what you can do. We have it in list form, but its not necessarily easy to stcan.

docs/guide/advanced/conditional-fields.md
abea: I'll try "connection." "association" and "relation" are also options.

I had in my head the case of an audience at a film festival or something like that.

abea:
I added a table for the special operators, of which we only have one so far. The "you can add more than one condition" thing doesn't fit in the table model. This sets us up for the future bits which will use the operators.

localghost443 (approved):

PR #66, opened 2021-04-14
PRO-920 users

abea:


docs/guide/users.md
```diff
@@ -0,0 +1,54 @@
+# Users and user roles
+
+A **"user"** in Apostrophe is an account (usually for a human or benevolent robot) that can be used to log into the website. Beyond that, users get specific sets of permissions based on the **role** they are assigned. More on that to come.
```
localghost443, line 3:
```suggestion
A **"user"** in Apostrophe is an account (usually for a human or benevolent robot) that can be used to log into the admin interface. Beyond that, users get specific sets of permissions based on the **role** they are assigned. More on that to come.
```

localghost443: or similar. I want to make sure its distinct from 'site user' (like commerce for example)

docs/guide/users.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Adding users with the CLI task
+
+The CLI task for adding users is primarily useful to add the very first admin user. The command is:
+
+```bash
+node app @apostrophecms/user:add my-user admin
+```
+
+The two final arguments passed to the task are, in order, the new username (`my-user`) and the user role (`admin`).
```
localghost443, line 17: I think we should role right out of the gate with a user table. Here might be good. Or maybe even above.

docs/guide/users.md
```diff
@@ -0,0 +1,54 @@
...
+![A page editor interface showing the permission tab and "who can view this" field](/images/users-visibility.png)
+
+## User roles
+
+Users are assigned one of four **user roles**. Each role is assigned a set of permissions that cover typical content editing, review, and administration duties.
+
+| User role | What can they do? |
+| --------- | ----------------- |
+| **Guest** | Guest users can log into the website and view content with their visibility set to "Login required." They cannot edit any content or view unpublished content. |
+| **Contributor** | Contributors may create, edit, and archive content, including the global doc. They cannot publish anything that does not publish automatically (e.g., images, files). They cannot take any action on users. |
```
localghost443, line 52: they can't archive images or files either. When they archive stuff it needs approval.

docs/guide/users.md
```diff
@@ -0,0 +1,54 @@
...
+| **Editor** | Editors have all the permissions of contributors. They can also *publish* content. They cannot take any action on users. |
+| **Admin** | Administrators have permissions to create, edit, archive, and publish any content. They are the only role that may manage users, including resetting passwords. |
```
localghost443, line 54: I think we should be a little more clear on what actions they can take on the users (creating, changing roles, etc)

localghost443 (changes requested):

docs/guide/users.md
abea: I removed the bit about archiving. I was mistaken that they could do that.

docs/guide/users.md
abea: I don't think roles should come before adding new users. And it's a lot to put in the middle of what are otherwise short sections. I'll add links down to the roles section.

docs/guide/users.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Adding users through the UI
+
+Once you have that first user, you have the option to add new ones through the user interface. The first and most important thing to know is that **only admin users can create and manage other users.** The admin user created with the CLI command above will have this ability.
+
+Open the users manager interface by clicking the "Users" button in the admin menu.
+
+![The admin menu at the top of the browser window with the Users button highlighted](/images/users-admin-bar.png)
+
+When the manager opens, click the "New User" button to open a fresh content editor modal. Populate the required fields, including the user's "Display Name," [role](#user-roles), and username. Also fill in the "Password" field so they can log into the website and save.
```
localghost443, line 27:
```suggestion
When the manager opens, click the "New User" button to open a fresh content editor modal. Populate the required fields, including the user's "Display Name," [role](#user-roles), username, and password.
```

localghost443: login and save is confusing.

docs/guide/users.md
```diff
@@ -0,0 +1,54 @@
...
+
+![A user editor modal with values filled in for our user, "Sam Wilson"](/images/users-editor.png)
+
+## Logging into an Apostrophe website
+
+Users can log into Apostrophe websites at the `/login` URL path for the website. If the website base URL (homepage) is `https://example.rocks`, the login page will be `https://example.rocks/login`.
+
+![The Apostrophe login page with username and password fields](/images/users-login.png)
+
+### Restricting content access
```
localghost443, line 37: This feels like it really is more related to.....document visibility than it is Permissions. Weird feeling. Either way, I think the heading should be more related to the actual option (Visibility). Since this doesn't exist within the permissions/user tab, I think this should be in pieces. idk.

docs/guide/users.md
```diff
@@ -0,0 +1,54 @@
...
+![A page editor interface showing the permission tab and "who can view this" field](/images/users-visibility.png)
+
+## User roles
+
+Users are assigned one of four **user roles**. Each role is assigned a set of permissions that cover typical content editing, review, and administration duties.
+
+| User role | What can they do? |
+| --------- | ----------------- |
+| **Guest** | Guest users can log into the website and view content with their visibility set to "Login required." They cannot edit any content or view unpublished content. |
+| **Contributor** | Contributors may create and edit content, including the global doc. They cannot publish anything that does not publish automatically (e.g., images, files). They cannot take any action on users. |
```
localghost443, line 52: They also can't archive content.

localghost443 (approved):

docs/guide/users.md
abea: It will get crowded if we list everything each group cannot do. I'm doing that in specific cases that seem most important to point out. In this case, contribs can add a bunch of stuff, but they can't publish. They _can_ discard drafts. If we talk about not being able to archive content, we should also talk about being able to discard drafts.

If we want to be exhaustive we could make a big table with every permission and a ✔️  or ❌ for each user role. I think our UI already does that in the user editor, so I don't think it's necessary. But if you do, then I'd rather do that than list it out exhaustively like this.

localghost443: @abea Good point. lets leave for now, but we can address this later if we feel the need.

localghost443 (approved):

PR #68, opened 2021-04-19
PRO-921 Adds page on the global doc

abea:
Short and sweet.


boutell (approved):

docs/guide/global.md
```diff
@@ -0,0 +1,59 @@
+# The global doc
```
localghost443, line 1:
```suggestion
# Global settings
```

localghost443: I get that this is a doc, but talking about it as a doc in the headline is inside baseball a bit. I know that real programmers will get it, but there is a middle ground.

docs/guide/global.md
```diff
@@ -0,0 +1,59 @@
...
+
+Editable settings used across the Apostrophe app can be managed through the **global doc**. The global doc is a special piece: it is created automatically and is the only doc of its type allowed. It is configured through the `@apostrophecms/global` piece type module.
```
localghost443, line 3:
```suggestion
Editable content or settings used across your app can be managed through the **global doc**. The global doc is a special piece: it is created automatically and there is only one. It is configured through the `@apostrophecms/global` piece type module.
```

docs/guide/global.md
```diff
@@ -0,0 +1,59 @@
...
+
+There are many good uses for the global doc, including configuring:
```
localghost443, line 5: I feel like this could be more of an intro, like "Global setting" -> example.

docs/guide/global.md
```diff
@@ -0,0 +1,59 @@
...
+
+- company information or navigation in the website footer
+- a special ID for a third-party service, such as Google Analytics
+- social media account URLs
+
+These are all piece of information that are used across many web pages and they may potentially change at some point. They also only need to be defined once or they at least serve as a site-wide default.
```
localghost443, line 11:
```suggestion
```

docs/guide/global.md
```diff
@@ -0,0 +1,59 @@
...
+
+To demonstrate, we may want to display an organizational Github URL in the website footer and a "Contact" page. We can add that to the global doc as we would on any other piece type.
```
localghost443, line 13: Social media, while more evil, is probably better...

docs/guide/global.md
```diff
@@ -0,0 +1,59 @@
...
+        label: 'Basics',
+        fields: ['githubUrl']
+      }
+    }
+  }
+}
+```
+
+::: note
+For normal piece types, we talk about adding fields to the *piece type*. Since there is always only one doc for the "global piece type," we'll often simply talk about configuring *the global doc* instead of *the global piece type*.
```
localghost443, line 36:
```suggestion
```
I don't think this is necessary. I think its only needed because we're overly defining it as a 'piece'. Its good to know that its technically a piece, but not to the extent where we need notes?

docs/guide/global.md
```diff
@@ -0,0 +1,59 @@
...
+:::
+
+Once the global doc has fields available it will be available for editing in the admin bar with other content types.
```
localghost443, line 39:
```suggestion
Once the global doc has been configured, the fields will be available for editing in the admin bar.
```

docs/guide/global.md
```diff
@@ -0,0 +1,59 @@
...
+  <footer>
+    {% if data.global.githubUrl %}
+      <a href="{{ data.global.githubUrl }}">Our Github organization</a>
+    {% endif %}
+  </footer>
+{% endblock %}
+```
+
+::: warning
+It can be easy to abuse the global doc. Since the global doc is available in every template it can be tempting to add all kinds of data to it. This can eventually lead to performance problems since **the whole global doc is fetched on every page request**. If it contains a lot of [content areas](/guide/areas-and-widgets/) and has [relationships](/guide/relationships.md) with other docs, that one request can get very large.
```
localghost443, line 56:
```suggestion
Since the global doc is available in every template it can be tempting to add all kinds of data to it. This can eventually lead to performance problems since **the whole global doc is fetched on every page request**. If it contains a lot of [content areas](/guide/areas-and-widgets/) and has [relationships](/guide/relationships.md) with other docs, that one request can get very large.
```

I would reduce this

localghost443 (changes requested):

localghost443 (changes requested):

localghost443 (approved):

PR #71, opened 2021-04-27
PRO-1239 Adds radio page

abea:
I never added a page for the radio field. I think because it's new in A3. It's now on my list, but this link is currently 404.


abea:
Switching to a draft. I'll bang out the actual radio page instead.

abea:
I copied the select page, changed the word "select" and changed the example. They work the same way.

localghost443 (approved):

PR #72, opened 2021-04-28
Adds a section on using widget data to the player section

abea:
This came up in Discord and I realized it was a gap in the section. This is how we use widget data in the player JS.

### DIFF
--- a/docs/guide/areas-and-widgets/custom-widgets.md
+++ b/docs/guide/areas-and-widgets/custom-widgets.md
@@ -105,7 +105,7 @@ Here are some two-column styles for people following along.
 
 When adding client-side Javascript for widget interaction, add a widget "player" to contain that code. The player will run only when the widget is used. It will also run when the editable area of the page is refreshed during editing.
 
-We can use the example of a basic collapsible section widget, `collapse-widget`. It will hide detail text until a user clicks the header/button.
+We can use the example of a basic collapsible section widget, `collapse-widget` (also known as an "accordion" or "disclosure" widget). It will hide detail text until a user clicks the header/button.
 
 ::: details Example collapsible widget code
 **Module configuration**
@@ -203,15 +203,15 @@ apos.util.widgetPlayers.accordion = {
 };
 ```
 
-[Credit goes to Heydon Pickering](https://inclusive-components.design/collapsible-sections/) for the accessible collapser example.
+[Credit goes to Heydon Pickering](https://inclusive-components.design/collapsible-sections/) for the accessible collapsible example.
 
 ### Using widget data in players
 
-These widget players do not know anything about the data in the widget. If we want to use widget data in your player code, we will need to provide it.
+Widget players do not have direct access to any *widget data*. If we want to use widget data in the player, we need to pass it in.
 
-On the other hand, template files *do* have access to the data. One good way to use data in a widget player is to insert it as a data attribute value in the template. The player can then look for that data attribute.
+Template files, On the other hand, *do* have access to widget data (they are rendered on the server). One good way to use data in a widget player is to insert it as a data attribute value in the template. The player can then look for that data attribute.
 
-For example, we could change our collapser widget code to include a possible `color` field value:
+For example, we could change our collapse widget to include a `color` field value:
 
 ```django
 {# modules/collapse-widget/views/widget.html #}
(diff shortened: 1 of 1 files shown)


boutell (commented):
This is good, you might want to mention you can use `| jsonAttribute` to safely push more complex data in an attribute and then use `JSON.parse` in the player. (Using `dump` does not cut it, it's not designed for escaping attributes.)

boutell (approved):

boutell (approved):

docs/guide/areas-and-widgets/custom-widgets.md
```diff
@@ -204,3 +204,42 @@ apos.util.widgetPlayers.accordion = {
 ```
 
 [Credit goes to Heydon Pickering](https://inclusive-components.design/collapsible-sections/) for the accessible collapser example.
+
+### Using widget data in players
+
+These widget players do not know anything about the data in the widget. If we want to use widget data in your player code, we will need to provide it.
```
localghost443, line 210: This is phrased a bit strangly, but I'm also hesitant to provide a suggestion. This is what I think its saying.

```suggestion
By default, widget players don't have access to widget data. Widget data must be explicitly provided to players in order to utilize it.
```

docs/guide/areas-and-widgets/custom-widgets.md
```diff
@@ -204,3 +204,42 @@ apos.util.widgetPlayers.accordion = {
...
+
+On the other hand, template files *do* have access to the data. One good way to use data in a widget player is to insert it as a data attribute value in the template. The player can then look for that data attribute.
```
localghost443, line 212:
```suggestion
On the other hand, template files *do* have access to widget data. One good way to use data in a widget player is to insert it as a data attribute value in the template. The player can then look for that data attribute.
```

docs/guide/areas-and-widgets/custom-widgets.md
```diff
@@ -204,3 +204,42 @@ apos.util.widgetPlayers.accordion = {
...
+
+For example, we could change our collapser widget code to include a possible `color` field value:
```
localghost443, line 214:
```suggestion
For example, we could change our collapser widget to include a `color` field value:
```

localghost443: @abea I'm not sure if this is worth pointing out, but the technical term for this is a disclosure widget. It has a complete spec, which accessibility tree, etc. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

docs/guide/areas-and-widgets/custom-widgets.md
```diff
@@ -204,3 +204,42 @@ apos.util.widgetPlayers.accordion = {
...
+
+```django
+{# modules/collapse-widget/views/widget.html #}
+<section data-collapser data-color="{{ data.widget.color }}" class="collapser">
+  {# The rest of the code is the same... #}
+</section>
+```
+
+We've added the `data-color` attribute to the widget wrapper with our color data. Then in the player code we could
```
localghost443, line 223:
```suggestion
We've added the `data-color` attribute to the widget wrapper with our color data. Then in the player code we could:
```

localghost443 (changes requested):

docs/guide/areas-and-widgets/custom-widgets.md
abea: 😬 I don't think I meant to end the sentence here.

docs/guide/areas-and-widgets/custom-widgets.md
```diff
@@ -203,4 +203,43 @@ apos.util.widgetPlayers.accordion = {
...
 ```
 
-[Credit goes to Heydon Pickering](https://inclusive-components.design/collapsible-sections/) for the accessible collapser example.
+[Credit goes to Heydon Pickering](https://inclusive-components.design/collapsible-sections/) for the accessible collapsible example.
+
+### Using widget data in players
+
+Widget players do not have direct access to any *widget data*. If we want to use widget data in the player, we need to pass it in.
+
+Template files, On the other hand, *do* have access to widget data (they are rendered on the server). One good way to use data in a widget player is to insert it as a data attribute value in the template. The player can then look for that data attribute.
```
localghost443, line 212:
```suggestion
Template files on the other hand, *do* have access to widget data (they are rendered on the server). One good way to use data in a widget player is to insert it as a data attribute value in the template. The player can then look for that data attribute.
```

localghost443 (changes requested):

localghost443 (approved):

PR #74, opened 2021-05-04
Beta routes

abea:
Updates docs with the columns ordering info and route `before` option.

### DIFF
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -17,7 +17,6 @@ Module configuration objects may use the following configuration properties. The
 | [`filters`](#filters) | Object | Yes | Configure piece type filters | Piece |
 | [`columns`](#columns) | Object | Yes | Configure piece type manager columns | Piece |
 
-
 ### "Cascading" settings
 
 Many Apostrophe module sections are structured as objects with `add`, `remove`, `group`, and `order` properties. This pattern allows these settings to "cascade" from the base classes to project level classes without requiring those settings be declared again.
@@ -88,7 +87,6 @@ modules.export = {
 ```
 
 You should not use both `improve` and `extend` in a single module. If "improving" an existing module, that existing module has already taken care of the "extending."
-
 
 ### `options`
 
@@ -468,7 +466,6 @@ Information returned by the component function will be available in the associat
 
 See the [async component guide](/guide/async-components) for more usage information.
 
-
 ```javascript
 // modules/product/index.js
 module.exports = {
@@ -679,7 +676,7 @@ Each extended API route function should accept the original function as `_super`
 
 ### `renderRoutes(self)`
 
-Add custom API routes to return rendered templates. The `apiRoutes` function takes takes the module as an argument and must return an object with properties for the relevant [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods), including `get`, `post`, `patch`, and `delete`. Each of those properties should be set to an object of functions.
+Add custom API routes to return rendered templates. The `renderRoutes` function takes takes the module as an argument and must return an object with properties for the relevant [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods), including `get`, `post`, `patch`, and `delete`. Each of those properties should be set to an object of functions.
 
 The name of the route dictates the template file that will be rendered. For example the `latest` route in the `product` module's `renderRoutes` section will return the template at `/modules/product/views/latest.html`.
 
@@ -726,7 +723,7 @@ We recommend using `apiRoutes` or `restApiRoutes` whenever possible before using
 // modules/product/index.js
 module.exports = {
   // ...
-  renderRoutes(self) {
+  routes(self) {
     return {
       get: {
         // GET /api/v1/product/redirect/:_id
--- a/docs/reference/module-api/module-overview.md
+++ b/docs/reference/module-api/module-overview.md
@@ -758,7 +758,7 @@ A route in the `apiRoutes`, `restApiRoutes`, `renderRoutes`, and `routes` sectio
 | Property | What is it? |
 | -------- | ----------- |
 | `route` | A route function that accepts a request as its argument |
-| `before` | Identify the middleware from a particular module. Used if the route function should execute before particular middleware runs. Format: `middleware:nameOfModule` |
+| `before` | Identify a module's middleware or routes. Used if the route function should be registered before particular middleware or other routes. Format as `middleware:nameOfModule` for middleware or `nameOfModule` for routes. |
 
 ### `handlers(self)`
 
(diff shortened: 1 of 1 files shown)


docs/reference/module-api/module-overview.md
```diff
@@ -527,24 +545,9 @@ To maintain the same application, they should return the same type of response a
 
 ### `restApiRoutes(self)`
 
-Add a custom REST API for a module. The `restApiRoutes` function takes the module as an argument and returns an object of functions that map to the standard REST API request types. Route functions may be asynchronous (async).
+Add a custom REST API for a module. The `restApiRoutes` function takes the module as an argument and returns an object of properties that map to the standard REST API request types.
 
-If you simply wish to add to the existing behavior of the REST API routes, see [`extendRestApiRoutes`](#extendrestapiroutes-self).
```
abea, line 532: I reordered a few of these to be consistent, so it looks like more changed than really was.

docs/reference/module-api/module-overview.md
```diff
@@ -566,6 +569,24 @@ module.exports = {
 };
 ```
 
+If including a route option, use an object. See the [route options](#route-options) section for more.
+
+If you simply wish to add to the existing behavior of the REST API routes, see [`extendRestApiRoutes`](#extendrestapiroutes-self).
+
+::: warning
```
abea, line 576: From here down in this green section didn't change.

docs/reference/module-api/module-overview.md
```diff
@@ -720,6 +741,25 @@ module.exports = {
 };
 ```
 
+If including a route option, use an object. See the [route options](#route-options) section for more.
+
+Each route function takes the Express arguments `req` (the [request object](https://expressjs.com/en/api.html#req)) and `res` (the [response object](https://expressjs.com/en/api.html#res)). The functions must generate a response via `res` to avoid leaking resources, typically using the `res.redirect` or `res.send` methods.
```
abea, line 746: From here down, until the `### Route options`, in this green section didn't change.

docs/reference/module-api/module-overview.md
```diff
@@ -566,6 +569,24 @@ module.exports = {
...
+- `getAll`
+- `getOne`
+- `post`
+- `patch`
+- `put`
+- `delete`
+
 #### `extendRestApiRoutes(self)`
 
 Extend the behavior of existing REST API routes in `extendRestApiRoutes`. This function must return an object of functions. See [`restApiRoutes`](#restapiroutes-self) for the valid function names.
```
boutell, line 592: Some of these accept an `_id` argument as well as `req`.

docs/reference/module-api/module-overview.md
```diff
@@ -689,23 +714,19 @@ module.exports = {
...
 
 ```javascript
 // modules/product/index.js
 module.exports = {
   // ...
-  renderRoutes(self) {
+  routes(self) {
     return {
       get: {
         // GET /api/v1/product/redirect/:_id
```
boutell, line 732: There's no `:_id` here. That sounds like a copy and paste from restApiRoutes which is the only place that parameter would ever be implied. As written this route would ned to get any arguments from `req.query`. However you could write:

```
'redirect/:_id': async (req, res) => { ... }
```

If you wanted a URL-based parameter to populate in `req.params._id`.

docs/reference/module-api/module-overview.md
```diff
@@ -720,6 +741,25 @@ module.exports = {
...
+:::
+
+### Route options
+
+A route in the `apiRoutes`, `restApiRoutes`, `renderRoutes`, and `routes` sections (as well as their `extend` variations) may include options to govern particular behavior. If so, it should be configured as an object rather than a function. That object should include a `route` property in addition to the options:
+
+| Property | What is it? |
+| -------- | ----------- |
+| `route` | A route function that accepts a request as its argument |
+| `before` | Identify a module's middleware or routes. Used if the route function should be registered before particular middleware or other routes. Format as `middleware:nameOfModule` for middleware or `nameOfModule` for routes. |
```
boutell, line 761: I would say "Identify another module's middleware or routes," I think it adds clarity about what this is intended to do.

boutell (changes requested):

docs/reference/module-api/module-overview.md
```diff
@@ -603,6 +626,8 @@ module.exports = {
 };
 ```
 
+If including a route option, use an object. See the [route options](#route-options) section for more.
```
boutell, line 629: Should include a guide to the error names that can be thrown, if I'm not missing it somewhere, which is very possible.

docs/reference/module-api/module-overview.md
```diff
@@ -689,23 +714,19 @@ module.exports = {
 };
 ```
 
-### `routes(self)`
+If including a route option, use an object. See the [route options](#route-options) section for more.
 
-Add standard Express routes. The `routes` function takes takes the module as an argument and must return an object with properties for the relevant [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods), including `get`, `post`, `patch`, and `delete`. Each of those properties should be set to an object of functions. Route functions may be asynchronous (async).
```
boutell, line 694: It is worth mentioning here that in most cases you will encounter fewer bugs if you use `apiRoutes` or `renderRoutes`, depending on your use case, and that classic Express routes are most useful if you need to use `res.redirect` or pipe a stream to `res`, two situations in which simply converting return values to responses and exceptions to status codes would be insufficient.

docs/reference/module-api/module-overview.md
abea: I'll note to do this in a separate PR. I'm not sure how we would want to handle that since it applies to other scenarios (rignt?).

docs/reference/module-api/module-overview.md
abea: We do have that note lower down: https://github.com/apostrophecms/a3-docs/pull/74/files#diff-65c9af67a92177dac44970f7f4e84f6d2a574d58e64e7b017a208bc3137d634eR750-R752

docs/reference/module-api/module-overview.md
```diff
@@ -758,7 +765,7 @@ A route in the `apiRoutes`, `restApiRoutes`, `renderRoutes`, and `routes` sectio
 | Property | What is it? |
 | -------- | ----------- |
 | `route` | A route function that accepts a request as its argument |
-| `before` | Identify a module's middleware or routes. Used if the route function should be registered before particular middleware or other routes. Format as `middleware:nameOfModule` for middleware or `nameOfModule` for routes. |
+| `before` | Identify another module's middleware or routes. Used if the route function should be registered before particular middleware or other routes. Format as `middleware:nameOfModule` for middleware or `nameOfModule` for routes. |
```
boutell, line 768: How about "registered before a particular module's middleware or routes." Routes are not "other" middleware.

boutell (changes requested):

boutell (approved):

PR #84, opened 2021-05-11
PRO-923 Updates the async components page

abea:
I think we should actually move this into the advanced section. It's not necessary for 90% of Apostrophe development and could be intimidating to new blood.


boutell (approved):
This is great. I think it should stay in the main flow of things — yes there are users who won't need it at first, but it's a necessary counterpart to keep people from just banging their heads on trying to put async functions in the helpers section, inject their own functions into templates for that purpose, etc. I've seen a lot of flailing over the years.

abea:
I think that's a good argument for clearly documenting this, which we didn't do as well for A2. We can point to it from the initial guide in an additional note block, but the first section is really about making people feel confident in starting to build with Apostrophe. The second section is for building on that foundational understanding, which is what this does.

PR #86, opened 2021-05-13
PRO-889 Adds hosting requirements page to new sidebar section

abea:


docs/guide/hosting.md
```diff
@@ -0,0 +1,46 @@
...
+
+We recommend hosting Apostrophe in self-hosted production environments with the following minimum specifications:
+
+| Software | Minimum Version |
+| ------------- | ------------- |
+| Ubuntu | 16+ |
+| [NGINX](https://www.nginx.com/) (or another proxy server) | Latest |
+| [Node.js](https://nodejs.org/en/) | 12.x+ |
+| [MongoDB](https://docs.mongodb.com/manual/administration/install-community/)  | 3.6+ |
+| [ImageMagick](https://imagemagick.org/script/download.php#macosx) (optional) | Any |
```
boutell, line 13: A macosx link doesn't make sense here since this is production. Also it would be better to always guide people to pages that encourage installation via the package manager. ImageMagick looks like the main offender there because that would be downloading a binary RPM and manually installing it, which creates a maintenance burden. 

This should be sufficient on ubuntu, needs testing:

```
sudo apt update && sudo apt install imagemagick
```

The other links you gave should work out better for folks.

docs/guide/hosting.md
```diff
@@ -0,0 +1,46 @@
...
+
+::: note
+MongoDB requires a minimum of 5GB free disk space at all times. Small websites may not need much more additional space than that. We have found that 20GB is sufficient for most Apostrophe applications, but more space may be necessary to handle an especially large number of file uploads or especially large uploaded files.
+:::
+
+## Deployment basics
+
+Deployment processes will vary depending on the hosting environment, technical requirements, and team preferences. There are a few steps that all deployment processes should include.
+
+1. If Apostrophe has been running on the server already, **stop the application process(es).**
```
boutell, line 31: This does not have to be done yet.

docs/guide/hosting.md
```diff
@@ -0,0 +1,46 @@
...
+2. **Run the data migration task: `node app @apostrophecms/migration:migrate`.** This will run both core data migrations as well as project-level migrations.
+3. **Run the build process: `node app @apostrophecms/asset:build`.** This runs automatically in development environments, but not if `NODE_ENV=production` or if the `autoBuild: false` option is set on the application. That is to ensure that the build is an intentional part of deployment and that developers can execute it along with other production build tasks as necessary.
+4. **Start (or restart) the application process(es).**
```
boutell, line 34: This is where you stop the old instance and start the new one. No need for the site to be down during the build of the next asset generation (which is slow) and the migrations (people who have exceptions to that rule will know).

docs/guide/hosting.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Best practices
+
+### Run multiple processes
+
+Running the website on multiple server processes is always a good idea in the production environment. It is usually not necessary to run more than two processes. Also make sure not to run more processes than you have CPU cores available (see your hosting plan).
```
boutell, line 40: You should run at least two, even if you have only one core, to keep the site available if one process is restarting. Otherwise no more than the number of cores.

docs/guide/hosting.md
```diff
@@ -0,0 +1,46 @@
...
+
+We recommend using a utility such as [PM2](https://pm2.keymetrics.io/) to start and run these processes. PM2 will also restart the processes in the rare case of a crash.
+
+### Minifying assets
+
+Apostrophe concatenates [project-level front end code](/guide/front-end-assets.md#placing-client-side-code) and, in browsers with a logged-in user, delivers it to browsers with the user interface code as well. There is no advanced minification applied. If needed, apply any advanced minification prior to running the production build task.
```
boutell, line 46: "in browsers with a logged-in user" is a little confusing to me. How about "if a user is logged in."

boutell (changes requested):
Looking good, a few notes.

docs/guide/hosting.md
```diff
@@ -0,0 +1,46 @@
...
+
+1GB of RAM may be sufficient if using a remote database service (e.g., MongoDB Atlas). Sites running in a multicore configuration will require additional 2G or more RAM and, of course, CPU cores.
+
+::: note
+MongoDB requires a minimum of 5GB free disk space at all times. Small websites may not need much more additional space than that. We have found that 20GB is sufficient for most Apostrophe applications, but more space may be necessary to handle an especially large number of file uploads or especially large uploaded files.
+:::
+
+## Deployment basics
+
+Deployment processes will vary depending on the hosting environment, technical requirements, and team preferences. There are a few steps that all deployment processes should include.
```
boutell, line 29: NODE_ENV should be set to `production` at all times in production and production-like environments (such as staging servers). This ensures that performance enhancements are enabled.

docs/guide/hosting.md
abea: That was from copying from the setting up page. I should remove it there as well.

docs/guide/hosting.md
abea: I don't understand. If you shouldn't have more processes than you have cores, why can we say to use two even if you have one core? Is it that running two on one core is worth the risk?

docs/guide/hosting.md
```diff
@@ -28,19 +28,23 @@ MongoDB requires a minimum of 5GB free disk space at all times. Small websites m
...
 ## Best practices
 
+### Set the `NODE_ENV` variable for production
+
+Apostrophe includes performance enhancements when in "production mode." In production environments, or staging environments that replicate production, set the `NODE_ENV` to `production`. Be aware that this will disable the automatic user interface build. See the section above for more about running the build process.
+
 ### Run multiple processes
 
-Running the website on multiple server processes is always a good idea in the production environment. It is usually not necessary to run more than two processes. Also make sure not to run more processes than you have CPU cores available (see your hosting plan).
+Running the website on multiple server processes is always a good idea in the production environment. We recommend running at least two. Also make sure not to run more processes than you have CPU cores available (see your hosting plan).
```
boutell, line 44: Sorry, but we do have to disambiguate a little here still. I don't want people to think they must have two cores, that's a bit expensive and not a requirement. How about:

"You should run at least two processes to guarantee availability if one process is restarting, even if you only have one CPU core. If you have more than two CPU cores, you may run additional processes, one per additional core. If you have a lot of capacity, you might want to reserve a core for MongoDB."

boutell (changes requested):

docs/guide/hosting.md
abea: works for me.

boutell (approved):

PR #91, opened 2021-05-20
PRO-915 Upgrading page

abea:


.styleguide.md
```diff
@@ -17,6 +17,12 @@ Bad:
 
 In most cases, punctuation should not be used at the end of titles. If the title is a question, exclamation, or quote, it would end in a question mark, exclamation point, or quotation mark, respectively.
 
+### "Levels," "types," and other wording style
+
+- When used as a compound adjective, "project-level" should be hyphenated (e.g., "project-level code"). When used as a noun referring to that layer of work, "project level" is not hyphenated (e.g., "add the file at project level") referring to work done in a project codebase, is not hyphenated. English is weird. Generally avoid the latter usage.
+- "Content type" is not hyphenated.
+- "Front end" and "back end" are not hyphenated. The dev world will never stop debating this. We have... for now.
```
abea, line 24: fwiw, after checking for the other two items, I realized that "front end" should simply change dependent on its context. So "I write front-end code" but "I work on the front end." "Front end developer" seems to be the question, which does seem correct without a hyphen since the developer themself is not "front-end."

Thank you for coming to my TED Talk.

agilbert (approved):

PR #94, opened 2021-05-26
PRO-964 Adds template data page

abea:


agilbert:
@abea can we get this deployed to a3.docs.apos.dev or can you tell me how I can preview these changes?

agilbert (changes requested):
My take on this page is that it would be worth revisiting the template type indicators. I'm a bit confused by the relationship of "Page" and "Show page" in the way it's described.

I'm wondering if it could be something like:

- all pages
- show pages only
- widgets

If you use these types, then it's always either "all pages" or "show pages only".

Maybe easier to discuss in real time.

agilbert (changes requested):
Please change the language:

"Page" > "All Pages"
"Show page" > "Show Pages"
"Widget" > "Widgets"

agilbert (approved):

PR #95, opened 2021-05-26
PRO-902 Makes additional clarification regarding where to put logo files

abea:
After looking around at other CMSes, and thinking about what we have to say about static media files, I decided it didn't make sense to have a full page about them or shoehorn it into a tangentially related media page. I added some additional direction in our existing public directory reference at the beginning instead.


docs/guide/technical-overview.md
```diff
@@ -17,7 +17,7 @@ There are a few directories and top-level files that are especially important in
 | ------ | ------ |
 | `app.js` | The heart of the application. This is where you tell Apostrophe what modules are in your project and set a few top-level parameters. |
 | `/modules` | All project-level [modules](/reference/glossary.md#module) and configuration for installed modules. |
-| `/public` | Public, static files. Apostrophe will generate specific directories inside, but you can also use it as needed.  |
+| `/public` | Public, static files (e.g., logo image files). Apostrophe will generate specific directories inside, but you can also use it as needed.  |
```
agilbert, line 20: I'm not sure this is the best example as there are plenty of instances where a website's logo image might be managed by the CMS. Could this just say, "e.g. any media assets you don't want managed by the CMS" ?

agilbert (changes requested):

agilbert (approved):

PR #97, opened 2021-06-03
Updates the minimumPark option reference

abea:


docs/reference/module-api/module-options.md
```diff
@@ -477,8 +477,7 @@ module.exports = {
 
 ### `minimumPark`
 
-The `minimumPark` option offers the ability to overwrite how the default required pages are configured. This includes the home page and the trash "page" (trash is part of the page tree). Most "parking" pages should be done with the [`park`](#park) option.
-<!-- TODO: Update when the trash is removed from home page children. -->
+The `minimumPark` option offers the ability to overwrite how the default required pages are configured. This includes the home page and the archive "page" (the archive is part of the page tree). Most "parking" pages should be done with the [`park`](#park) option.
```
boutell, line 480: At this point I would replace the whole paragraph with:

"This option sets the initial defaults for the mandatory "Home" and "Archive" parked pages. You should leave this option at its default setting and, if desired, reconfigure the home page as the first entry in your `park` option, taking care to set `parkedId` to `home` just like `minimumPark` does. Reconfiguration of `minimumPark` would make sense only in an npm module intended to change the defaults for all sites that use it."

boutell (changes requested):

docs/reference/module-api/module-options.md
abea: nevermind. I'm beefing up the warning.

boutell (approved):

PR #100, opened 2021-06-08
adds note about passing `format`

stuartromanek:


docs/reference/field-types/color.md
```diff
@@ -38,6 +38,12 @@ themeColor: {
 <!-- |contextual | Boolean | false | If `true`, it will prevent the field from appearing in the editor modal | -->
 <!-- |readOnly | Boolean | false | If `true`, prevents the user from editing the field value | -->
 
+### Options
+General options you can pass to a `color` field.
+| Option | Type | Default | Description |
+|---------------|------|---------|-------------|
+| `format` | String | `hex8` | The color string format to be saved. Possible options are "rgb", "prgb", "hex6", "hex3", "hex8", "name", "hsl", "hsv". |
```
abea, line 45: this should be part of the "Color picker options" table below, right?

abea (changes requested):

docs/reference/field-types/color.md
abea: I'm not opposed to changing how it works in core. As it is currently built those picker options are documented below, including a code example showing how to nest it.

abea: It looks like this is related to the core PR https://github.com/apostrophecms/apostrophe/pull/3122/. I'll respond with more there.

abea: Actually, see the area field page for how we're already doing this. The format is different, giving more space to the option.

abea (changes requested):

docs/reference/field-types/color.md
stuartromanek: expanded options and pickerOptions sub props

abea (approved):

PR #102, opened 2021-06-15
Adds CLI install docs and CLI tips

abea:


boutell (approved):
This is a good improvement. We should work on clarifying which steps a person doesn't have to do if they have used the cli, like npm install.

PR #104, opened 2021-06-16
PRO-1607 Updates header to switch to V2 docs

abea:


agilbert (commented):
@abea these look good to me though I'm not sure that the `themeConfig` part is for and how to review that. Is it possible to see the changes on the staging site and approve from there?

docs/.vuepress/config.js
```diff
@@ -39,25 +39,49 @@ module.exports = {
     logo: '/images/apos-dark.png',
     nav: [
       {
-        text: 'Guide',
-        link: '/guide/introduction.md'
+        text: 'v2 Docs',
+        link: 'https://v2.docs.apostrophecms.org'
```
agilbert, line 43: Actually, can we make this a "Versions" dropdown so you can see there are the two versions here?

agilbert (changes requested):

agilbert (approved):

PR #108, opened 2021-06-18
PRO-1627 PRO-1625 styles

stuartromanek:
- adds code syntax theme
- adds code copy button
- formats code boxes
- styles headings
- adds ordinal heading type
- adds styleguide page as sandbox
-

### DIFF
--- a/docs/.vuepress/theme/styles/type.styl
+++ b/docs/.vuepress/theme/styles/type.styl
@@ -6,6 +6,7 @@ $type-size-h1 ?= 2.62rem;
 $type-size-h2 ?= 2.62rem;
 $type-size-h3 ?= 2rem;
 $type-size-h4 ?= 1.56rem;
+$type-size-h5 ?= 1.31rem;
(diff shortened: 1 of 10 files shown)


docs/.vuepress/theme/styles/type.styl
```diff
@@ -0,0 +1,54 @@
...
+
+h1, h2, h3, h4, h5 {
+  font-weight: 700;
+  letter-spacing: -0.75px;
+  border-bottom: none;
+}
+
+h1 {
+  font-size: $type-size-h1;
+  a;
```
abea, line 28: This is crashing the build

abea (changes requested):
Looks good other than the stray `a;`.

There are two places we have H5s in use. They look okay with this but smaller than body text. Might be worth putting that on the same size at least.

docs/.vuepress/theme/styles/type.styl
stuartromanek: fixed

PR #113, opened 2021-07-01
Removes children page api query param

abea:
It looks like this was either never a thing or was removed. We may want to replace it with a depth option.


boutell:
Actually this is a real thing, but it looks like it's buggy. We turn on children() initially in getRestQuery, but then we call applyBuildersSafely, and the children query builder has a launder method, so that should accept the input you mentioned.

The problem is that the launder method doesn't seem capable of accepting any string other than the empty string as "false":

```
          launder(input) {
            let value = null;
            if (input) {
              if ((typeof input) === 'object') {
                value = {};
                if (input.depth) {
                  value.depth = self.apos.launder.integer(input.depth);
                }
              } else {
                value = true;
              }
            }
            return value;
          },
```

I think this should be updated so that if `input` is not an object, it gets passed through `apos.launder.boolean`, rather than just tested for truthiness, which will be true for all nonempty strings. Then the documented behavior would be accepted.

`ancestors` on the other hand lacks a launder method altogether, and we call it with `true` in `getRestQuery` so it is always on and can't be shut off. A launder method for this one would also be good, with the same structure.

PR #115, opened 2021-07-06
PRO-1638 headless

abea:


docs/guide/headless-cms.md
```diff
@@ -0,0 +1,177 @@
+# Using Apostrophe as a headless CMS
+
+Apostrophe is fully featured as a traditional CMS (with some non-traditional features). It is also immediately ready for use as a headless CMS. There are many scenarios where Apostrophe works well as a headless CMS, including:
```
agilbert, line 3: Apostrophe is fully featured as a traditional CMS. It is also immediately ready for use as a headless CMS. There are many scenarios where Apostrophe works well in a decoupled architecture, including:

docs/guide/headless-cms.md
```diff
@@ -0,0 +1,177 @@
...
+Before going further, keep in mind that there is an [API route reference section](/reference/api/) that will include much of the following information in a more direct format. One section there that will not be fully covered in this guide will be the [authentication APIs](/reference/api/authentication.md). There are a few options for authentication discussed there.
+
+For the sake of simplicity we can use the example of authentication using an API key. The key could be defined in the `@apostrophecms/express` module in a project like this:
+
+```javascript
+// modules/@apostrophecms/express/index.js
+module.exports = {
+  options: {
+    apiKeys: {
+      // Use your own strong, randomly generated key. Not like this one.
```
agilbert, line 22: Should we include a note about how these keys could also be stored as environment variables rather than in code? (Assuming that's possible...)

docs/guide/headless-cms.md
```diff
@@ -0,0 +1,177 @@
...
+
+Apostrophe is fully featured as a traditional CMS (with some non-traditional features). It is also immediately ready for use as a headless CMS. There are many scenarios where Apostrophe works well as a headless CMS, including:
+
+- We may want to use Apostrophe to serve webpages normally, but also use the website data in another context, such as a mobile app.
+- We may want to go fully headless with a front end like Gatsby or Gridsome while also allowing editors to use Apostrophe's in-context editing.
+- We may have an app that handles features such as e-commerce but need a CMS to create and pipe content into the app.
+
+Regardless of the reason or the layers involved there are some common concepts to understand.
+
+## Accessing data using the REST APIs
```
agilbert, line 11: Should we include a note that we don't have GraphQL support yet, but it's something we're considering for our roadmap? (We should probably actually prioritize this on our roadmap...)

agilbert: [You can let folks know to share how important this is to them on our roadmap.](https://portal.productboard.com/apostrophecms/1-product-roadmap/c/44-graphql-api)

docs/guide/headless-cms.md
```diff
@@ -0,0 +1,177 @@
...
+
+```
+GET https://example.rocks/api/v1/@apostrophecms/page?flat=1&all=1apikey=000apikey123
+```
+
+## Getting rendered HTML in responses
+
+Using an object of JSON-like data from a REST API is very normal from a headless CMS. Traditionally the application's "head" (such as a single page Vue.js, React, or Angular application) would use that structured data to display product, event, or similar content information. It is less common for a headless CMS to deliver page content that is tied to particular HTML markup and layout. More often, landing page text and images might be coded directly into the app code or otherwise not be from the same data source.
+
+Apostrophe support both approaches. The in-context editing features and [widget areas](/guide/areas-and-widgets.md) give editors more control over content, but in it's default form areas are fairly incomprehensible data objects. It is when they are rendered with the widget template files that they are fully realized.
```
agilbert, line 151: Apostrophe supports both approaches. The in-context editing features and [widget areas](/guide/areas-and-widgets.md) give editors more control over content, but in its default form areas are fairly incomprehensible data objects. It is when they are rendered with the widget template files that they are fully realized.

agilbert: I corrected some basic grammar/typo stuff above. I think the description of this could use some reworking in future iterations, but OK with this for now.

docs/guide/headless-cms.md
```diff
@@ -0,0 +1,177 @@
...
+```
+GET https://example.rocks/about-us
+```
+
+ More likely we would only want the unique page content (without the `head` tag and outer wrapper elements), which we can get by adding the `aposRefresh=1` query parameter. It will include only the HTML markup for the page template.
+
+```
+GET https://example.rocks/about-us?aposRefresh=1
+```
+
```
agilbert, line 170: Can we add a note that describes examples of when you might use these different approaches?

agilbert (changes requested):
Some relatively minor changes requested in here, but it's a good start overall.

docs/guide/headless-cms.md
abea: Not as ENV variables, but they could be in data/local.js. That's covered more in the authentication page that I linked in the first paragraph in this section.

We should have more documentation of how to use environment configurations. I've created a ticket for that.

docs/guide/headless-cms.md
abea: 😬 I'm not sure how much we should be doing that in docs. I guess we could use a note block to point to product board for input.

abea: I've added that at the end.

agilbert (approved):

PR #118, opened 2021-07-14
PRO-963 Adds template tag reference page

abea:


docs/reference/README.md
```diff
@@ -3,4 +3,5 @@
 - [Glossary](/reference/glossary.md): Terminology and concepts that are important or useful to know when working with ApostropheCMS
 - [Field Types](/reference/field-types/): Core schema field types and their options
 - [REST API and additional API routes](/reference/api/): Built-in API routes for working with Apostrophe site content
-- [Module API](/reference/module-api/): Reference material for building and configuring Apostrophe modules
\ No newline at end of file
+- [Module API](/reference/module-api/): Reference material for building and configuring Apostrophe modules
+- [Template tags](/reference/template-tags): Apostrophe-specific template tag that add additional functionality in templates.
```
boutell, line 7: tag -> tags

docs/reference/template-tags.md
```diff
@@ -0,0 +1,191 @@
...
+    <div>
+      {{ rendercaller() }}
+    </div>
+  </section>
+{% endfragment %}
+```
+
+### Arguments
+
+#### `name()`
```
boutell, line 113: What's with the parens?

boutell: OK I see, it's because the whole declaration is an argument to the fragment tag. I think it would be clearer to me if "params..." appeared inside the parens or something.

docs/reference/template-tags.md
```diff
@@ -0,0 +1,191 @@
...
+The fragment name used to reference it in `render` and `rendercall` tags. It should be written as a function with parentheses. The parentheses may include argument names if needed, which would be referenced in variable brackets.
+
+## render
+
+The `render` tag is used to insert a [fragment](/guide/fragments.md) in a template.
+
+### Usage
+
+```django
+{% render name(arguments) %}
```
boutell, line 124: Maybe use another word to avoid confusion between the arguments to this tag and the arguments to the fragment?

boutell: I accept that name(params...) is a single "argument" to the nunjucks tag, I'd just suggest another word (like params) be used to describe what's passed to the fragment, so there's less confusion between the levels.

boutell (changes requested):

docs/reference/template-tags.md
```diff
@@ -0,0 +1,191 @@
+# Template tags
+
+Template tags add additional functionality to Apostrophe templates. The tags described below are specific to Apostrophe, though they use the standard Nunjucks syntax: `{% tagName %}`. See standard tags in the [Nunjucks reference](https://mozilla.github.io/nunjucks/templating.html#tags).
```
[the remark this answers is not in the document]
abea: I'll spice it up a bit.

docs/reference/template-tags.md
```diff
@@ -0,0 +1,191 @@
...
+
+The primary argument is a combination of the name of a module name and the name of an async component from that module, separated by a colon. All async components belong to a specific module, though multiple modules may have components with the same name.
+
+#### `data` (optional)
+
+The data argument, following the `with` keyword, is available in the async component template as `data`. It can be any data type, however it is a best practice to make it an object with subproperties so using the values are clearer in templates.
+
+## fragment
+
+The `fragment` tag *declares* a template fragment that will be inserted elsewhere with [`render`](#render) or [`rendercall`](#rendercall). See the [template fragments guide](/guide/fragments.md) for more on using this feature.
```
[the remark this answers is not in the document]
abea: We say that very explicitly in the fragments guide, which I linked to there. I want to avoid additional diversions in the reference section as much as possible.

docs/reference/template-tags.md
```diff
@@ -0,0 +1,191 @@
...
+{% fragment name(arguments) %}
+  {# Fragment markup #}
+{% endfragment %}
+```
+
+**Examples:**
+```django
+{% fragment button(text, action, options = {}) %}
+  <button class="o-button {{ options.class }}" data-action="{{ action }}">
+    {{ text }}
```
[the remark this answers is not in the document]
abea: I don't really think that's the job for this reference section. The guide goes into that as well. The card example with `rendercaller()` does connect to the `rendercall` tag example that uses an area tag. But we also want to establish this as the standard method for synchronous cases as well.

docs/reference/template-tags.md
abea: I'm not sure I understand. The arguments here and those in the fragment function are the same. they go in here and out there.

abea: Are you saying we need a term inside both tags' parentheses or that we need _different_ terms in each term's parentheses?

abea: My understanding is that [arguments go in, parameters come out.](https://developer.mozilla.org/en-US/docs/Glossary/Parameter) I'll start by updating it that way.

docs/reference/template-tags.md
```diff
@@ -0,0 +1,200 @@
...
+    sizes: '(min-width: 600px) 45vw, (min-width: 1140px) 530px'
+  }
+} %}
+```
+
+### Arguments
+
+#### `context`
+
+The area's document context, either a page (`data.page`), piece (`data.piece`), or widget (`data.widget`). The area field must be defined in the field schema for that context. See the [template data](/guide/template-data.md) section for more on each `data` property.
```
boutell, line 44: Or an array item.

docs/reference/template-tags.md
```diff
@@ -0,0 +1,200 @@
...
+The name (a string) of the area field as defined in the field schema.
+
+#### `contextOptions` (optional)
+
+The context options object is added after area tag arguments following the `with` keyword. It is an object with keys matching the names of widget types allowed in that particular area field. Each key is assigned a value that will be passed into the widget template as `data.contextOptions`. See the [image widget guide](/guide/core-widgets.md#image-widget) for an example.
+
+Context options are optional for all core and official Apostrophe widget types.
+
+::: note
+Context options are not the best place for most widget configuration. That should be done in the [area field configuration](/reference/field-types/area.md#widgets). Context options are used to override that with options that only apply to the specific template context.
```
boutell, line 57: I would say "supplement," not "override." You can't override anything. (OK technically if a particular widget.html is specifically written to look in one place before the other you could "override" in that sense, but we can't suggest in any way that you can override the widget schema of the area or people will keep trying to do that 😂)

So I would change "override" to "supplement" and add this sentence:

"Context options cannot  change which widgets are permitted in an area."

docs/reference/template-tags.md
```diff
@@ -0,0 +1,200 @@
...
+    </div>
+  </section>
+{% endfragment %}
+```
+
+### Arguments
+
+#### `name(parameters)`
+
+The fragment name used to reference it in `render` and `rendercall` tags. It should be written as a function with parentheses. The parentheses may include argument names if needed, which would be referenced in variable brackets.
```
boutell, line 124: How about: "It should be written *like* a function declaration, with parentheses."

I don't understand "which would be referenced in variable brackets," I suggest just removing that.

Since this is reference you should probably add "parameters should be separated by commas and may optionally include a default value assignment."

boutell (changes requested):
Sorry to come back so late.

abea:
Thanks! Much better.

docs/reference/template-tags.md
```diff
@@ -49,14 +53,18 @@ The context options object is added after area tag arguments following the `with
...
+
+## `component`
 
 The `component` tag inserts an asynchronous component into the template. See the [async components guide](/guide/async-components.md) for more on using this feature.
 
 ### Usage
 
 ```django
-{% component 'moduleName:componentName' with data %}
+{% component 'module:componentName' with data %}
```
boutell, line 67: Not sure why we wouldn't want consistency, we don't say `component`, why do we say `module`?

boutell (approved):

docs/reference/template-tags.md
abea: In the case of `componentName`, we're in the section about `component` and already used that term in this example (as the tag keyword). I'm not sure what my thinking was at the time, but I'm guessing I used `module` so this would be less verbose. It is explained below as well.

PR #126, opened 2021-08-20
PRO-883 PRO-1757

abea:
I've put all this on the staging docs site for easier reading. Please add comments here as normal: https://a3.docs.apos.dev/guide/localization/

Dynamic l10n is coming.


.stylelintrc
```diff
@@ -8,6 +8,7 @@
   ],
   "rules": {
     "at-rule-disallowed-list": ["each", "function"],
+    "scale-unlimited/declaration-strict-value": "never",
```
abea, line 11: We're mixing our SCSS stylelint rules with stylus in Vuepress. Gotta clean that up at some point.

agilbert (commented):
I found these to be very readable and clear from my perspective, awesome work - the code samples are really shining on this one.

One thing I noticed was that when I was on the localization page, it didn't seem to highlight in the sidebar and automatically show the children to that page. I find this to be helpful wayfinding in moving around through the docs. Is it possible to do that?

abea:
@agilbert There's an option to make the section not collapsible at all. I think that might be the best we could do out of the box. Could be appropriate here, but if we start doing that a bunch the sidebar will get crowded fast.

We could instead highlight the sub pages more prominently on the landing page than the links I have now.

docs/guide/localization/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+::: note What about "internationalization"?
+The term "internationalization," often abbreviated **i18n** comes up when discussing this as well. And for good reason. See [the localization section of the glossary](/reference/glossary.md#localization-terms) for a quick overview of how we use these terms. Put simply, we use "internationalization" to refer to the system that supports the localization processes. That's why the core Apostrophe modules is `@apostrophecms/i18n`.
+:::
+
+## Configuring locales
+
+The first step in localizing content, whether static or dynamic, is to define locales. These are most often different languages, countries, or combinations of the two. In some cases locales it may also be appropriate to establish locales for people based on topical interests, professional categories, or cultural identities.
```
boutell, line 16: Grammar: "in some cases locales it may also..."

docs/guide/localization/README.md
```diff
@@ -0,0 +1,135 @@
...
+- a `baseUrl` must be set on the application, defining the default hostname, OR
+- a `hostname` setting must be used on all locales
+
+For this example we'll assume we have `baseUrl: 'example.com'` set on the application.
+:::
+
+So how does Apostrophe choose the best locale to use? In many cases it is clear. If there is conflict, however, the best locale uses the following prioritization:
+1. The locale matches *both* the URL's `hostname` *and* path `prefix`.
+2. The locale matches the URL's `hostname` only.
+3. The locale matches the URL's path `prefix` only.
```
boutell, line 118: "The URL matches the locale's configured `prefix` and the locale has no `hostname`."

docs/guide/localization/README.md
```diff
@@ -0,0 +1,135 @@
...
+::: note
+If `hostname` is used for any locale:
+- a `baseUrl` must be set on the application, defining the default hostname, OR
+- a `hostname` setting must be used on all locales
+
+For this example we'll assume we have `baseUrl: 'example.com'` set on the application.
+:::
+
+So how does Apostrophe choose the best locale to use? In many cases it is clear. If there is conflict, however, the best locale uses the following prioritization:
+1. The locale matches *both* the URL's `hostname` *and* path `prefix`.
```
boutell, line 116: "The locale has both a `prefix` and a `hostname` and the URL matches both."

docs/guide/localization/README.md
```diff
@@ -0,0 +1,135 @@
...
+2. The locale matches the URL's `hostname` only.
+3. The locale matches the URL's path `prefix` only.
+4. The locale is higher in the configuration object.
+5. The locale is the default locale (when no other locale matches).
```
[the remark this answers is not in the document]
boutell: In addition, *if any locales have a hostname, then they all must, or the top-level `baseUrl` option of Apostrophe must be set.*

docs/guide/localization/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+## The default locale
+
+The default locale is the locale used when no others matche the URL better. It is typically the locale used by your website's primary audience. **If no locales are configured, Apostrophe will use `'en'` as the default locale name.**
```
boutell, line 124: matche -> match

docs/guide/localization/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+The default locale can be changed in one of the following ways:
+- Configuring one locale (ideally the first) with *no* `prefix` or `hostname` properties
```
boutell, line 127: If `defaultLocale` is not set, then the first configured locale acts as the default, even if it has a `hostname` and `prefix`. It's best to set `defaultLocale` for avoidance of confusion. The `defaultLocale` is always used for requests associated with command line tasks unless the `locale` command line argument is present.

docs/guide/localization/README.md
```diff
@@ -0,0 +1,135 @@
...
+- Making sure no lower-specificity locales appear earlier in the `locales` configuration
+- Using [the `defaultLocale` option](/reference/modules/i18n.md) on the `@apostrophecms/i18n` module to name another locale
+
+::: warning
+It is important to configure locales, especially the default locale, before content entry begins. This is especially the case if your locales will not include `'en'`. If content entry begins on the default `'en'` locale and we later configure locales that do not include `'en'`, that original content will disappear.
```
boutell, line 132: will seem to disappear until `en` is added back.

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
...
+        {% endfor %}
+      </ul>
+    </section>
+  ```
+  <template v-slot:caption>
+    /modules/article-page/views/show.html
+  </template>
+</AposCodeBlock>
+
+Registering that string is as easy as wrapping it in a template method: `__t()`. Make sure the string passed to the method is in quotes and any matching quotes in the string are escaped (e.g., `__t('Gritty\'s friends')`).
```
boutell, line 28: template helper, or template helper function when it's not relaxing with friends. It's not being invoked with method syntax, so I think that language is confusing.

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
+# Static content localization
+
+This section covers localization of strings throughout template files and Apostrophe user interface. These strings are hard-coded, not editable by logged-in users, and thus are not stored in the database. Instead, this localization is stored in JSON files associated with the various locales.
+
+## Registering strings to localize
```
boutell, line 5: I would just say "localizing strings." "Register" implies that it somehow knows it's registered once you invoke `__t` and will, for instance, populate `en.json` for you, which is not the case. You're invoking the helper and localizing the string right here, you're not registering it for later.

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
...
+  </template>
+</AposCodeBlock>
+
+Registering that string is as easy as wrapping it in a template method: `__t()`. Make sure the string passed to the method is in quotes and any matching quotes in the string are escaped (e.g., `__t('Gritty\'s friends')`).
+
+<AposCodeBlock>
+  ```django
+    {# More article template stuff above ⤴ #}
+    <section>
+      <h2>__t('Related articles')</h2> {# 🎉 It's registered! #}
```
boutell, line 34: I would say "it's localized."

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
...
+  module.exports = {
+    // ...
+    apiRoutes(self) {
+      return {
+        get: {
+          // GET /api/v1/product/newest
+          async newest(req) {
+            const product = await self.find(req).sort({ createdAt: -1 }).toObject();
+            if (!product) {
+              // 👇 The 404 error is registered with req.t()
```
boutell, line 74: Sends a localized version of the 404 error message with `req.t`

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
...
+<AposCodeBlock>
+  ```javascript
+  module.exports = {
+    // ...
+    apiRoutes(self) {
+      return {
+        get: {
+          // GET /api/v1/product/newest
+          async newest(req) {
+            const product = await self.find(req).sort({ createdAt: -1 }).toObject();
```
boutell, line 72: Nice example.

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
...
+  <template v-slot:caption>
+    /modules/product/index.js
+  </template>
+</AposCodeBlock>
+
+::: tip
+There is currently no browser method available to register strings in project client-side JS. A good alternative is to create an API route that returns localized strings then use the response from that route in client-side JS.
+:::
+
+If you are customizing Apostrophe **user interface components**, the `$t()` method is available in the component (used as `this.$t()`) outside the template block.
```
boutell, line 94: It's available in the template block too, as just `$t`. This language makes it sound like it's not available there.

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
...
+
+::: note
+As a reminder, the Vue.js components of the user interface are not connected to any Vue app you may be running for your website visitors. The registration method will not be automatically available outside the UI components.
+:::
+
+## Adding and using localization files
+
+Now that there are strings to localize, we need to add JSON files for locales with the strings and their translations. For the purposes of our example we will assume we have two locales: `'en'` (English) and `'es'` (Spanish).
+
+Each locale should get its own JSON file using the locale name, in a `i18n` directory of a module that has internationalization active. Configure that module with the [`i18n` option](/reference/module-api/module-options.md#i18n), such as:
```
boutell, line 106: Subdirectory.

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
...
+
+Like all modules, we would need to activate this in `app.js`. Our JSON files would be:
+
+- `modules/localization/i18n/en.json`
+- `modules/localization/i18n/es.json`
+
+::: tip
+These files can be spread across modules if you like. For example you might want to localize strings from a particular module's template files in that module directory. Like with many things, this depends how you and your team like to work.
+
+It's important to note that **if the same key is used in multiple template files they will be rendered to the same string**. It does not matter if each template has a localization JSON file with that key registered.
```
boutell, line 129: People really want to know where to put stuff, and they will put it in `@apostrophecms/i18n` because that sounds right. But they should not put their own phrases there, unless it's for the apostrophe admin UI, because that module is already using the `apostrophe` namespace. So we should explicitly note that you should not put your project level phrases in that module at project level.

docs/guide/localization/static.md
boutell: I get what you're trying to do with this word, and "localizing" by itself also has the wrong implication since we're just using what you localized in a json file somewhere. How about "displaying localized strings"?

docs/guide/localization/static.md
boutell: Better: "Displays the localized string!"

docs/guide/localization/static.md
```diff
@@ -0,0 +1,199 @@
...
+<h2>__t('myTeam:relatedArticles')</h2>
+```
+
+Apostrophe will then treat keys with the namespace differently from the same key without the namespace (`myTeam:relatedArticles` vs. `relatedArticles`). If someone uses the version *without* the namespace it will not overwrite the version *with* the namespace.
+
+## Localizing the Apostrophe user interface
+
+The Apostrophe user interface contains many registered strings, currently localized to English. We will be working to provide more localization files, but if you are interested in adding l10n files to your project for the UI, you are welcome to do that.
+
+Add JSON files for the locales as normal in the project-level `modules/@apostrophecms/i18n` module directory. It's important to use this directory since it already has the `'apostrophe'` namespace configured. The UI keys all use this namespace. In each JSON file, copy the contents of [the Apostrophe core l10n file](https://github.com/apostrophecms/apostrophe/blob/main/modules/@apostrophecms/i18n/i18n/en.json) to get all the keys. You can then start translating each string.
```
boutell, line 197: You're welcome to use this directory, because it already has the namespace configured, but I wouldn't say it's important. In fact, if you're thinking of sharing your translations in npm for others to use, you'll be happiest starting off with your own module and setting `ns: 'apostrophe'` there, which multiple modules are welcome to do.

docs/reference/glossary.md
```diff
@@ -123,6 +128,22 @@ The editing interface for this join allows the user to pick offices to associate
 <!-- See the [apostrophe-schemas](/reference/modules/apostrophe-schemas/README.md) module documentation for more information about joins. -->
 
 <!-- TODO: Display this when other locale information is added. -->
-<!-- ## Document set
+<!-- ### Document set
+
+The database documents that exist as mode and locale versions of one another. For example, in a project with an English and a French locale, the home page would be managed in the database by six documents: 1) the draft English, 2) published English, 3) previous English, 4) draft French, 5) published French, and 6) previously published French. These six database documents make up the document set. -->
```
boutell, line 133: previously published English (consistency)

boutell: You could mention that they all share an `aposDocId`, which is also a prefix of the `_id` of each document.

docs/reference/module-api/module-options.md
```diff
@@ -125,6 +126,34 @@ module.exports = {
 }
 ```
 
+### `i18n`
+
+If set to a truthy value, Apostrophe will look for an `i18n` directory *in that module* with one or more JSON files with localization string definitions (e.g., `modules/project-languages/es.json`). The JSON file names should match configured locale names. [See the static i18n guide](TODO) for more information.
```
boutell, line 131: subdirectory

docs/reference/module-api/module-options.md
```diff
@@ -125,6 +126,34 @@ module.exports = {
...
+
+If set to an object, there may be `ns` and `browser` properties.
+
+| `i18n` setting | Value type | Description |
+|---------|---------|---------|
+| `ns` | String | A namespace for localization string keys in this module. If undefined, Apostrophe will use the `'default'` namespace. That namespace is intended for project-level localization. |
```
boutell, line 137: The namespace `'apostrophe'` is reserved for Apostrophe's UI. You may intentionally set `ns` to `'apostrophe'` if your goal is to localize the Apostrophe user interface.

docs/reference/modules/i18n.md
```diff
@@ -0,0 +1,98 @@
+# `@apostrophecms/i18n`
+
+**Alias:** `apos.i18n`
+
+This module governs internationalization functionality and localization tools. Apostrophe projects will configure locales through `@apostrophecms/i18n`.
+
+The module makes an instance of the [i18next](https://npmjs.org/package/i18next) npm module available in Nunjucks templates via the `__t()` helper function. That function is also available on `req` objects as `req.t()` and in user interface Vue.js components as `this.$t()`. Any options passed to this module are passed on to `i18next`.
```
boutell, line 7: Nope — we do not pass on all options like we used to do with `i18n`. And that's a good thing because it gives us more control over what i18next might be doing that we do or don't support, etc. We can ease into allowing more options to be passed over time.

docs/reference/modules/i18n.md
```diff
@@ -0,0 +1,98 @@
...
+## Options
+
+|  Property | Type | Description |
+|---|---|---|
+|`defaultLocale` | String | The locale that will be used by the UI and server rendering if no other is specified. |
+|`locales` | Object | The set of [locales](../glossary.md#locale) to use in the application. |
+
+### `locales`
+
+The locale object should include one or more (usually two or more) locale configuration object. Each locale key is a short code, typically a two letter country code (e.g., `ca`), language code (`fr`), or one of each with a dash separating them (`ca-fr`). This local name is used to reference the locale throughout Apostrophe.
```
boutell, line 20: object -> objects

Your dashed example should be `ca-FR`

docs/reference/modules/i18n.md
```diff
@@ -0,0 +1,98 @@
...
+
+Each locale may have the following settings:
+
+|  Locale setting | Type | Description |
+|-----|-----|-----|
+| `label` | String | The human-readable label for the locale. |
+| `hostname` | String | A hostname (e.g., 'example.net') that will trigger the locale to be shown. |
+| `prefix` | String | A URL path prefix that will trigger the locale to be shown. |
+
+Hostname and the path prefix are both factors in deciding what local to display to visitors. There are a some factors that factor into identifying the correct one to use. The priority ranking for choosing the correct locale is:
```
boutell, line 30: "there are a some" -> grammar

docs/reference/modules/i18n.md
```diff
@@ -0,0 +1,98 @@
...
+
+1. The locale matches *both* the URL's `hostname` *and* path `prefix`.
```
boutell, line 32: See earlier notes and suggested language.

docs/reference/modules/i18n.md
```diff
@@ -0,0 +1,98 @@
...
+3. The locale matches the URL's path `prefix` only.
+4. The locale is the default locale (when no other locale matches).
+
+Other notes:
+
+- Two or more locales may not be registered with the same `hostname` *and* the same path `prefix`. Apostrophe will throw an error in this case.
+- The **default locale** (either the `defaultLocale` setting or the first registered locale) does not need a `hostname` or path `prefix` setting.
+- If *any* locale has a `hostname` setting *one* of these must be true:
+  1. **The Apostrophe app must have a `baseUrl` set in the `app.js` file** (a best practice in most cases anyway), OR
+  2. **All locales must have a hostname setting** (even if several are the same).
```
boutell, line 43: even if several are the same, distinguished by prefix.

docs/reference/modules/i18n.md
```diff
@@ -0,0 +1,98 @@
...
+- If the URL does not match any locale's set `hostname` or `prefix` (and all locales have one or both settings), Apostrophe will use the default hostname.
```
boutell, line 44: will use the default locale

boutell (changes requested):

docs/reference/glossary.md
abea: this bit isn't actually new (or being published. though this might be the time to revisit it.

docs/guide/localization/README.md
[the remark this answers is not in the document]
abea: I do mention this just prior to this list in a note.

PR #129, opened 2021-08-27
PRO-1634 Adds information about relationship fields

abea:


docs/guide/relationships.md
```diff
@@ -218,3 +218,44 @@ With this field in place, you could display connected articles in a topics show
...
+    },
+    // ...
+};
+```
+
+In the example above, we add a `fields` property to the relationship field, just like the `fields` property on the module itself. Just like the primary schema, we use an `add` subproperty with field configurations inside it. In this case there is no need for a `group` subproperty.
+
+Once this is added, editors can select an "Edit Relationship" option on the selected relationships.
+
+![The relationship field now with a menu button and "edit relationship" option](/images/relationship-fields.png)
```
agilbert, line 261: Nice. Is it possible to get a screenshot of what it looks like editing the relationship with the fields you've configured in this example?

agilbert (changes requested):
Minor request for an additional screenshot.

docs/guide/relationships.md
abea: added

agilbert (approved):

PR #130, opened 2021-08-31
Pro 1925 Nav recipes

abea:


abea:
It's two ways.

docs/cookbook/building-navigation.md
```diff
@@ -0,0 +1,292 @@
...
+      }
+    }
+  }
+  ```
+  <template v-slot:caption>
+    modules/@apostrophecms/page/index.js
+  </template>
+</AposCodeBlock>
+
+For example, **to include two levels of pages in the nav** we would add to the `builders` option that ancestors (such as the home page) should include two levels of child pages (`depth: 2`).
```
boutell, line 68: People will take this literally. Should say it makes it *possible* to include two levels of pages in the nav, and then show how to do that at template level.

docs/cookbook/building-navigation.md
```diff
@@ -0,0 +1,292 @@
...
+      }
+    }
+  }
+  ```
+  <template v-slot:caption>
+    modules/@apostrophecms/page/index.js
+  </template>
+</AposCodeBlock>
+
+By default the page data object includes all properties from the database document. We can **limit that returned data** (for a minor speed improvement and clearer logging) by adding a projection to the query builder. *Make sure to include the `path` in this projection to avoid crashing.*
```
boutell, line 91: Ouch! But true. Want to open a ticket to automatically include path in projections of pages or something?

docs/cookbook/building-navigation.md
```diff
@@ -0,0 +1,292 @@
...
+<AposCodeBlock>
+  ```javascript
+  module.exports = {
+    options: {
+      builders: {
+        children: true,
+        ancestors: {
+          children: {
+            depth: 2,
+            project: {
```
boutell, line 102: Projections are fairly tricky. Consider suggesting using `areas: false, relationships: false` as a simpler starting point. That gives at least 90% of the benefit.

docs/cookbook/building-navigation.md
```diff
@@ -0,0 +1,292 @@
...
+
+As you can see, page templates come ready with page tree data that is ready to become site navigation. With additional configuration we can customize the data that templates give us. The examples above are only a few such ways to configure it.
+
+## Add fields for manual nav building
+
+The other common approach to site navigation is to let editors build it manually. They would be able to choose the specific page links and the order in which the links should appear. Since the site nav is the same across every page (in most cases), we store the navigation data in [global settings](/guide/global.md).
+
+Before we look at a code example, let's think about how we would want to define fields for site navigation.
+
+- Website navigation is an order list of links, some of which might contain their own list of links (second-level navigation). This translates well to an array structure, so **we use an [`array` field type](/reference/field-types/array.md)**.
```
boutell, line 160: I would say "the" rather than "an" here. I'm sure this is the kind of review you needed from me, amirite?

boutell (changes requested):
Also please cover how to achieve a breadcrumb trail, as that introduces `data.page._ancestors`. You might also mention that any time `data.piece` is present, it's usually a good idea to treat it like an extra level for breadcrumb purposes.

docs/cookbook/building-navigation.md
abea: I'm trying not to overwhelm too much with examples, but I'll add a note about the additional layer of `_children` array in `data.home._children` items. And fix the misleading bold bit here.

docs/cookbook/building-navigation.md
abea: I'll do that.

docs/cookbook/building-navigation.md
abea: yep 😃 I think it was supposed to be "an array field." I'll fix that.

boutell (changes requested):
These improvements are good — still need to show use of `data.page._ancestors` in templates. Then I think we've covered the possibilities well.

docs/cookbook/building-navigation.md
```diff
@@ -0,0 +1,318 @@
+# Building site navigation
+
+Almost every website needs some kind of primary navigation. It's usually a row of links at the top of each page that is hidden behind some kind of "hamburger menu" button on smaller devices. For Apostrophe sites there are two main approaches to adding and maintaining site navigation:
```
agilbert, line 3: I think you can remove this sentence: "It's usually a row of links at the top of each page that is hidden behind some kind of "hamburger menu" button on smaller devices."

agilbert (changes requested):
Minor suggested edit, not a huge deal but just tightening up the content a bit.

agilbert (approved):

docs/cookbook/building-navigation.md
```diff
@@ -282,5 +282,37 @@ The final step is to turn the array data from this into template markup. We will
...
 
+**What is breadcrumb navigation, really?** If we think through the pieces of breadcrumb nav, we are looking at: *A series of links, usually starting with the home page, proceeding through the page tree, down to the page we are on.*
+
+`data.page._ancestors` perfectly matches this goal. It is an array of page objects, starting with the home page, continuing through the page tree and ending with the parent of the rendered page. Perfect! To turn that into a breadcrumb navigation, we simply need to **loop through `data.page._ancestors`, add a link for each ancestor**.
+
+The only tricky part comes at the end. If a visitor is on a piece's [show page](/guide/piece-pages.md#the-show-page-template), then `data.page` will actually be the index page. If `data.piece` exists (indicating we're rendering a show page), we will link to the index page (`data.page`) in the breadcrumbs, otherwise we'll simply print the current page's title.
```
boutell, line 293: You would still want to list the index page itself, as it is sensibly an "ancestor" of the piece. Otherwise the breadcrumb for most pieces would be:

`Home | Productname`

Rather than:

`Home | Products | Productname`

boutell (changes requested):

boutell (approved):
Never mind all good

PR #132, opened 2021-09-07
Matched method name

dharmesh-hemaram:
Updated method name in the template to match with one defined in index.js


abea:
Thank you, @dharmesh-hemaram!

abea (approved):

PR #135, opened 2021-09-09
Pro-1944 🆕 updates

abea:


docs/guide/headless-cms.md
```diff
@@ -12,22 +12,30 @@ Regardless of the reason or the layers involved there are some common concepts t
 
 Before going further, keep in mind that there is an [API route reference section](/reference/api/) that will include much of the following information in a more direct format. One section there that will not be fully covered in this guide will be the [authentication APIs](/reference/api/authentication.md). There are a few options for authentication discussed there.
 
-For the sake of simplicity we can use the example of authentication using an API key. The key could be defined in the `@apostrophecms/express` module in a project like this:
+One of the more common use cases for a headless CMS is to simply get data from the CMS for a public website built with React, Vue.js or some other framework. Since the certain data is publically available in the front-end application, we can simply make that data publically available directly through the REST API using [a `publicApiProjection` option](/reference/api/authentication.md#allowing-public-access) on the doc type. This will only apply to `GET` requests.
```
boutell, line 15: publically -> publicly

docs/guide/headless-cms.md
```diff
@@ -36,30 +44,29 @@ Requesting and using [piece](/guide/pieces.md) data should seem familiar to deve
...
-https://example.rocks/api/v1/article?apikey=000apikey123
+https://example.rocks/api/v1/article
 ```
 
 When using Apostrophe as a headless CMS, we will most frequently be making `GET` requests from the Apostrophe app (see the [pieces API reference](/reference/api/pieces.md) for more on other request types). There are two main types of `GET` requests we should think about: a "get all" request and a "get one" request.
 
 ### Get all the pieces!
 
-To make a **"get all"** request, hit the URL as described above. Using the native Node.js `fetch` utility, that might look like:
+To make a **"get all"** request, hit the URL as described above. Using the Node.js `node-fetch` utility, such a request in a front end application might look like:
```
boutell, line 54: In a "front-end application" you would not need node-fetch at all, right? You'd just use window.fetch (aka just use fetch).

boutell: Feels like we still need more context here. I think it might even be worth presenting a tiny little index.html vanilla js app that does the whole thing.

docs/guide/headless-cms.md
```diff
@@ -90,10 +96,10 @@ Page REST API requests look very similar to piece requests. They follow the same
 
 ```
 # All pages
-GET https://example.rocks/api/v1/@apostrophecms/page?apikey=000apikey123
```
boutell, line 93: The root query for pages behaves differently by default from the regular REST convention because pages are arranged in a tree, not a flat list. See below (copied from the code).

"Trees are arranged in a tree, not a list. So this API returns the home page,
with _children populated if ?_children=1 is in the query string. An editor can
also get a light version of the entire tree with ?all=1, for use in a
drag-and-drop UI.

If ?flat=1 is present, the pages are returned as a flat list rather than a tree,
and the `_children` property of each is just an array of `_id`s.

If ?autocomplete=x is present, then an autocomplete prefix search for pages
matching that string is carried out, and a flat list of pages is returned,
with no `_children`. This is mainly useful to our relationship editor.
The user must have some page editing privileges to use it. The 10 best
matches are returned as an object with a `results` property containing the
array of pages.

If querying for draft pages, you may add ?published=1 to attach a
`_publishedDoc` property to each draft that also exists in a published form."

boutell: The main thing I'm flagging here is that it doesn't get "all pages" by default.

docs/guide/headless-cms.md
```diff
@@ -153,7 +159,7 @@ Apostrophe supports both approaches. The in-context editing features and [widget
 **In any Apostrophe REST API single-document `GET` ("get one") request, include the `?render-areas=1` query parameter.** This will return the page document as JSON. Each area field in the response will have a `_rendered` property with a string of HTML instead of the `items` array of widget objects. This could be useful if we mostly want to insert the rendered area HTML into an existing template system.
 
 ```
-GET https://example.rocks/api/v1/@apostrophecms/page/903y5n76e8j:en:published?apikey=000apikey123&render-areas=1
+GET https://example.rocks/api/v1/@apostrophecms/page/903y5n76e8j:en:published?render-areas=1
 ```
 
 If we wanted to get the full page as HTML rendered using the page templates, we could make a `GET` request to the page URL (basically what browsers do). This is only good if we are not trying to insert it into other templates.
```
boutell, line 165: You can add `?aposRefresh=1` to just get the part of the page that is editable in Apostrophe, which could be useful in this situation. Further customization is possible using a custom query parameter and we need to cover that frequently used technique but that's clearly a scope for a new doc ticket.

docs/guide/media.md
```diff
@@ -32,28 +32,36 @@ module.exports = {
...
 ```django
-{# modules/article-page/views/show.html #}
 {% area data.piece, 'photo' %}
 ```
+  <template v-slot:caption>
+    modules/article-page/views/show.html
+  </template>
+</AposCodeBlock>
 
 Done. The core image widget is design to render a responsive image with alt text (if it was entered for the image). When that is all that's needed this is a great option.
```
boutell, line 55: is designed

docs/guide/media.md
```diff
@@ -67,91 +75,122 @@ module.exports = {
...
 
 <a href="{{ fileUrl }}">Download</a>
 ```
+  <template v-slot:caption>
+    modules/article-page/views/show.html
+  </template>
+</AposCodeBlock>
+
 
 For image attachments, doing the same thing will return the URL for the `full` image size by default. You can pass an options object as a second argument with its `size` property set to another image size to get a different URL back.
```
boutell, line 105: I think it should be mentioned that `full` is not the original size.

docs/guide/upgrading.md
```diff
@@ -8,29 +8,30 @@ Apostrophe 3 introduces a number of new features for developers and content-edit
 
 ### 100% RESTful API coverage
 
-In the previous version, developers who wanted REST API access to their data needed to install the `apostrophe-headless` module. [These are now built into A3](/reference/api/) and are the foundation of the user interface.
+In the previous version, developers who wanted REST API access to their data needed to install the `apostrophe-headless` module. [This is now built into A3](/reference/api/) and are the foundation of the user interface.
```
boutell, line 11: and is the foundation

docs/guide/upgrading.md
```diff
@@ -8,29 +8,30 @@ Apostrophe 3 introduces a number of new features for developers and content-edit
...
 
 ### Clearer and easier module building
 
-There were several things about building advanced modules in A2 that often tripped up developers. A3 helps avoid the confusion around where to put things by doing away with `beforeConstruct`, `construct`, and `afterConstruct`. [They are replaced by a single `init` function and special sections](/reference/module-api/module-overview.md#initialization-function) to do things like extend API routes, add template helper methods, and add columns to the piece manager UI.
+There were several things about building advanced modules in A2 that often tripped up developers. A3 helps avoid the confusion around where to put things by doing away with `beforeConstruct`, `construct`, and `afterConstruct`. [They are replaced by a single `init` function and special sections](/reference/module-api/module-overview.md#initialization-function) to do things like extend API routes, add template helper methods, and add columns to the piece manager UI. The `addFields` array also became a `fields` object with an `add` subproperty to prevent problems with common inheritance patterns.
```
boutell, line 19: `fields` can optionally be a customization function too, but OK, don't want to blow their minds too hard at this stage.

docs/guide/upgrading.md
```diff
@@ -56,25 +57,27 @@ Your project specific modules should not use the `@apostrophecms` namespace to a
...
   - `columns` (for piece manager UI columns) is structured similarly to `fields` with `add`, `remove`, and `order` subproperties. This replaces A2's `addColumns` property.
   - `columns` (for piece manager UI columns) is structured similarly to `fields` with `add`, and `remove` subproperties. This replaces A2's `addFilters` property.
-  - Many module options that were top level properties in A2 (e.g., `label`, `alias`, `sort`) are now subproperties of `options`. These include all options with primitive value (e.g., booleans, string, numbers) in addition to select object or array options. See the [module options reference](/reference/module-api/module-options.md).
+  - Many module options that were top level properties in A2 (e.g., `label`, `alias`, `sort`) are now subproperties of `options`. These include all settings that are *not* covered in the [module properties overview reference](/reference/module-api/module-overview.md). See the [module options reference](/reference/module-api/module-options.md) for more on that.
```
boutell, line 60: How about "Many module settings". In A3, strictly speaking, if it ain't in `options`, it ain't an option, it's something else.

docs/guide/upgrading.md
```diff
@@ -56,25 +57,27 @@ Your project specific modules should not use the `@apostrophecms` namespace to a
...
+  <!-- TODO: Update once options are all moved into individual module reference pages -->
   - The A2 `self.addHelper()` method used to add Nunjucks helper functions is replaced by the `helpers` section in module configuration.
```
boutell, line 62: This list is great. Looking forward to the links.

docs/guide/upgrading.md
```diff
@@ -56,25 +57,27 @@ Your project specific modules should not use the `@apostrophecms` namespace to a
...
+  - Custom `lib/cursor.js` files for modules are no longer loaded. Migrate this code to the new `queries` section as well.
+  - The A2 `self.expressMiddleware` system for adding Express middleware functions is replaced by the `middleware` section in module configuration.
   - The A2 `self.addTask()` method for adding CLI tasks is replaced by the `tasks` section in module configuration.
 
 ### Areas and pages
 
 - Every page type will need a corresponding module (in A2 this was only necessary if the page had custom fields or functionality). Page templates live in the page type module rather than in the base page module.
 - Areas must be declared as fields in the [content field schema](/guide/content-schema.md). They will no longer work if simply added to template files without being registered.
-- The area template tag format is now `{% area data.page, 'areaName' %}`. Note that this no longer includes the object of widget configuration since that is now done in module config.
+- The area template tag format is now `{% area data.page, 'areaName' %}`. Note that this no longer includes the object of widget configuration since that is now done in module configuration.
```
boutell, line 76: How about "no longer includes the configuration of widgets," seems a little needlessly technical the other way.

boutell (changes requested):

docs/guide/headless-cms.md
abea: that's covered in the next paragraph.

docs/guide/headless-cms.md
abea: True about node-fetch. Moving too quickly.

My goal here is to give the gist of how this works. We can open tickets to add more context, but I don't have something at hand to add as a full example and I think we need to be okay with done on this one for now.

boutell (approved):

PR #138, opened 2021-09-21
Additional notes on the i18n browser option

abea:


docs/guide/localization/static.md
```diff
@@ -122,6 +122,23 @@ Each locale should get its own JSON file using the locale name, in a `i18n` subd
   </template>
 </AposCodeBlock>
 
+**If the l10n strings will be used in the user interface**, set the `i18n` option to an object with `browser: true`. *This includes any piece type and page type labels* that will appear in the UI. If that is not present then the UI build process will ignore the l10n JSON file.
```
boutell, line 125: This is great, but it's worth adding that it's only relevant to the admin UI for editors and these phrases are currently never available browser-side to the public.

boutell (changes requested):

boutell (approved):

PR #149, opened 2021-11-09
PRO-2257 batch ops docs

abea:


docs/reference/module-api/module-overview.md
```diff
@@ -287,6 +288,64 @@ modules.export = {
 };
 ```
 
+### `batchOperations`
+
+Piece types can use batch operations in the `batchOperations` cascade object property, which are actions editors can take on many selected pieces at once. Apostrophe has archive and restore (from the archive) batch operations by default, for example. New batch operations are added to a series of buttons in the piece type manager modal.
```
boutell, line 293: How about "piece types can offer batch operations via the..."

docs/reference/module-api/module-overview.md
```diff
@@ -287,6 +288,64 @@ modules.export = {
...
+
+#### `add`
+
+The `add` property is an object containing batch operation configurations. Each batch operation configurations should include the following properties:
```
boutell, line 297: configurations -> configuration (second appearance)

docs/reference/module-api/module-overview.md
```diff
@@ -287,6 +288,64 @@ modules.export = {
...
+
+| Property | Description |
+| ------- | ------- |
+| `label` | A text label used for the batch operation button's readable label. |
+| `route` | A relative route path with leading slash where the manager modal should send the selected document IDs for processing. The route should be defined in [`apiRoutes`](#apiroutes-self). |
```
boutell, line 302: After reading this I don't know what this route should look like. `/api/v1/something`? `/archive` but it gets appended to the `action` of the module?

For consistency with elsewhere in A2 and A3, if it starts with `/` it should be treated as the full `path` component of the URL, bypassing `action`. If it doesn't, then it should get appended to `action`, with a `/` between. One of many examples:

```
`${apos.i18n.action}/locale`
```

I don't know if this is a documentation bug or an implementation issue the documentation is faithfully describing.

docs/reference/module-api/module-overview.md
```diff
@@ -287,6 +288,64 @@ modules.export = {
...
+
+```javascript
+// modules/article/index.js
+modules.export = {
+  batchOperations: {
+    add: {
+      reset: {
+        label: 'Reset',
+        // Uses a hypothetical route added in `apiRoutes`
+        route: '/reset',
```
boutell, line 318: Leading `/` in an apostrophe route should mean we're bypassing any automatic prefixing with a module action URL, for consistency with how we handle it elsewhere. Should be just `reset` and code adjusted to match.

boutell (changes requested):

docs/reference/module-api/module-overview.md
abea: It's an  implementation issue the documentation is faithfully describing.

boutell (changes requested):
We need to document what the route will receive. Ideally with an example route, which could be one of the batch routes already in core, so no new code there.

docs/guide/batch-operations.md
```diff
@@ -0,0 +1,222 @@
...
+<AposCodeBlock>
+  ```javascript
+  module.export = {
+    // `batchOperations` and other module settings...
+    apiRoutes(self) {
+      return {
+        post: {
+          reset(req) {
+            // Make sure there is an `_ids` array provided.
+            if (!Array.isArray(req.body._ids)) {
```
boutell, line 111: Use `self.apos.launder.ids` to make sure you wind up with strings in that array, avoiding crashes or extra sanitization steps for malicious input

(It's OK to still check if it's an array first if you want and throw invalid if it's not)

docs/guide/batch-operations.md
```diff
@@ -0,0 +1,222 @@
...
+
+            // The iterator function that updates each individual piece.
+            async function resetter (req, id) {
+              const piece = await self.findOneForEditing(req, { _id: id });
+
+              if (!piece) {
+                throw self.apos.error('notfound');
+              }
+
+              // 🪄 Do the work of resetting piece field values.
```
boutell, line 140: Seems to me we could do an actual thing here for additional clarity, but this is a big incremental improvement.

boutell (changes requested):

docs/guide/batch-operations.md
abea: Okay, we don't do that in our archive and restore routes. Is it not necessary or should we add it there? I recall you saying that since we have this coming from the Apos UI that extra launder wasn't necessary.

docs/guide/batch-operations.md
abea: If it were simply changing `archive` to a boolean that might be alright. I'm not aware of a way to reset only fields with default values in only a few lines. I'd like to keep the focus of this on what batch operations need to work.

docs/guide/batch-operations.md
boutell: Oof... if I said that I was mistaken. You can never trust the browser.

That said, it's not crazy urgent, it's only an unproven hypothetical that something other than an id string could cause a problem here and it certainly can't cause a security breach. I will open a ticket to follow up implementation-wise, if you'd please take care of it here in the docs example.

boutell (approved):
Thanks.

PR #152, opened 2021-11-22
Pro-2267 migration documentation

abea:
The PR is in draft state while upgrader tool work continues.

It should be ready for initial review, however.


docs/.vuepress/enhanceApp.js
```diff
@@ -3,6 +3,14 @@ export default ({ router }) => {
     {
       path: '/guide/module-format-example.html',
       redirect: '/reference/module-api/example.html'
+    },
+    {
+      path: '/guide/upgrading.html',
```
abea, line 8: Redirects to the new section from the old URL

docs/guide/migration/overview.md
```diff
@@ -0,0 +1,36 @@
+# Changes from Apostrophe 2
```
abea, line 1: This page is mostly unchanged from the original ["Coming from 2.x" ](https://v3.docs.apostrophecms.org/guide/upgrading.html) page. The breaking changes section is now in the upgrading.md file.

docs/README.md
```diff
@@ -20,7 +20,7 @@ ApostropheCMS is a full-featured, open source CMS built with Node.js that seeks
 - [**Getting started**](/guide/setting-up.md): Environment setup, installation, and an overview of the major version changes.
 - [**Guide**](/guide/introduction.md): Discussion of Apostrophe 3 features using lots of code examples.
 - [**Reference**](/reference/): Detailed reference sections for developers.
-- [**Coming from Apostrophe 2**](/guide/upgrading.md): A guide to feature and API changes from Apostrophe 2.
```
boutell, line 23: Is this an existing page with a changing URL? Is there a plan for redirects & no loss of SEO?

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+## Breaking changes
+
+The updates described below are the majority of significant breaking changes, but are not completely comprehensive.
+
+### Code syntax, APIs, and module configuration
+
+#### Codebase structure and module naming
+
+- The modules directory moved from `/lib/modules` to `/modules` since there was little need for the additional nesting.
+- All core modules have been namespaced with `@apostrophecms`. For example, the `apostrophe-pages` module is now `@apostrophecms/page`. You can configure this in projects in `/modules/@apostrophecms/page/index.js`.
```
boutell, line 18: "At project level, you can configure this module in..."

Otherwise it sounds like you can configure the namespacing.

boutell: I recognize this is probably old language, but worth tweaking.

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+- Areas must be declared as fields in the [content field schema](/guide/content-schema.md). They will no longer work if simply added to template files without being registered.
+- The area template tag format is now `{% area data.page, 'areaName' %}`. Note that this no longer includes the configuration of widgets since that is now done in module configuration.
+- Area "singletons" are no longer a separate field type or template helper. They were always simply areas that only allowed one widget. With the other area changes, there is not much benefit to having that feature over adding the `max: 1` option to an area field.
+
+#### Client-side assets
+
+- A2 widget player formats are deprecated. The "lean" widget player structure can be more easily converted to the [A3 widget player](/guide/custom-widgets.md#client-side-javascript-for-widgets) due to a similar structure.
+- The `pushAsset` method for delivering CSS and client-side JS is replaced by [placing files in the correct location](/guide/front-end-assets.md).
+- The `/public` directory *within modules* is [moved to `/ui/public`](/guide/static-module-assets.md).
```
boutell, line 58: This was never really correct, and now that modules can have a `public` subdir again which is a close match for what `public` did in 2.x, it's probably best just to delete this line.

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+- The `array` field type uses [a new syntax for adding its field schema](/reference/field-types/array.md#module-field-definition), matching the new module field schema syntax.
+- The `tags` field from A2 no longer exists. In most cases we recommend replacing that by adding a piece type for categorization. The core `@apostrophecms/image-tag` and `@apostrophecms/file-tag` modules are examples of this.
+- Template helper methods need to be adjusted to use the singular form of their module aliases, e.g., `apos.attachments.url` to `apos.attachment.url`.
+
+### Data Structure
+
+The following changes apply to database documents in A2 databases' primary content collection, `aposDocs`, unless otherwise specified.
+
+- Apostrophe core *document* type names (the `type` property) change to reflect the A3 version names (e.g., `apostrophe-image`, `@apostrophecms/image`).
```
boutell, line 71: The A3 module names. And also in general piece type names in the database are always equal to piece type module names in A3.

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+- Apostrophe core *widget* type names (the `type` property) change to reflect the A3 version names (e.g., `apostrophe-video` to `@apostrophecms/video`).
```
boutell, line 72: The `type` property is now always identical to the module name, minus the `-widget` suffix.

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+- There are always at least two database documents for each published document: one representing the published document and another representing the draft, potentially with unpublished changes. Once changes are saved after initial publication, another for the "previous" state is added. The "previous" document is not required for initial operation, however.
```
boutell, line 73: This statement is only true if a document type is localized, if a piece type has the `localized: false` option setting then there are no modes for it (users being the main A3 example, but it's useful for anything always synced from a third party system where localization and publication wouldn't make sense because there is no local editing anyway).

Also, the published version does not always exist. Only the draft is guaranteed to always exist.

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+- Unique ID and document state properties evolved to support the document version system.
+  - The `_id` property is now a combination of a unique string, a locale identifier (defaults to `en`), and the document mode (e.g., `ckokisysc00048d4l5zdo9l0c:en:published`).
+  - A new property, `aposLocale` stores the full locale, e.g., `en:published`.
+  - A new property `aposMode` stores the document mode, e.g., `published`.
+  - A new property `aposId` stores the unique document ID (shared between each variation), e.g., `ckokisysc00048d4l5zdo9l0c`.
```
boutell, line 78: `aposDocId`, not `aposId`.

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Migration tools and process
+
+Migrating an Apostrophe 2 codebase and data should be done with care, but there are two official tools that can help speed the process along. Both tools are still somewhat early in development, so it is important to use them locally (not in production) and make sure to back up code and data before using them.
+
+**The Apostrophe team is very interested in feedback on these upgrader tools.** With each of their limitations in mind, please provide any comments, bug reports, code contributions, or feature requests as issues on the respective Github repositories.
+
+### Content Upgrader
+
+While completely manual code migration is possible, the *Content* Upgrader tool is basically essential for data conversion. Writing custom data migrations would not be able to do many thing differently from the official tool and there is very little to gain by doing so.
```
boutell, line 97: many thing -> many things

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+
+The Content Upgrader tool does not change anything on the original Apostrophe 2 database. Instead, it reads that original database and creates a *new database* using a name the developer provides. That new database will contain the original data, converted for use in Apostrophe 3.
```
boutell, line 99: anything on -> anything in

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+
+Developers will install the Content Upgrader as a module within the A2 project. This allows it to access schemas and other important project information. **See the [Content Upgrader README](https://www.npmjs.com/package/@apostrophecms/content-upgrader) for full instructions.**
```
boutell, line 101: I would bold "within the A2 project."

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+
+- Users and groups are not migrated. This is because the user roles of A3 differ in design from the permissions groups of A2. Copying users directly could create security issues. Create new accounts on the A3 project or arrive at your own migration strategy.
+- A2 has a built-in `apostrophe-images` widget type that acts as a slideshow when multiple images are selected. A3 only has a built-in single-image `@apostrophecms/image` widget type. By default `apostrophe-images` will be upgraded to `@apostrophecms/image`, with only the first image present in each. However you can use the `mapWidgetTypes` option (see the README) to override this mapping during the upgrade.
+- Since A3 does not have a built-in cropping feature yet, there is currently no accommodation for it in A3's `@apostrophecms/image` widget type. The tool attempts to carry over the cropping data in the format which is expected to work in A3 in the near future.
+
+### Code Upgrader
+
+The *Code* Upgrader tool support upgrading codebases for **full Apostrophe projects** (websites) and **installable modules**. It has two major roles in converting an A2 codebase to use Apostrophe 3:
```
boutell, line 113: support -. supports

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+
+1. It will lint an Apostrophe codebase for A2 syntax and structure that must change. This is its `lint` command.
+2. It will *make* many of those code changes for you. This is its `upgrade` command.
+
+Additionally, there is a `reset` command that can undo the automated changes. Always use a new git branch during this process as well to have an additional way to roll back changes.
```
boutell, line 118: `reset` actually relies on git too and won't be able to limit its effects to your changes if you already have unrelated changes in your repo that aren't committed.

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+
+One reason we recommend using this tool to execute changes on a full project is that it will make minimal code adjustments for A3 use. This is important because additional changes (such as complete module name changes), could unnecessarily break compatibility with the [upgraded database](#content-upgrader). After running the automatic code upgrade, take care in the final code changes to avoid effecting data compatibility.
```
boutell, line 120: We're probably overstating the benefits here, and it's also not clear why the developer would make "complete module name changes."

effecting -> affecting

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+
+The Code Upgrader is installed globally in a Node environment (including developer environments) and run as a command line tool. **See the [Code Upgrader README](https://www.npmjs.com/package/@apostrophecms/code-upgrader) for full instructions.**
+
+#### Limitations
+
+Just as with the content tool, the Code Upgrader has some limitations to understand. While database structure is very predictable, code styles and patterns are not. Some A2 APIs and syntax are intentionally not touched to avoid making incorrect assumptions.
+
+- The tool is designed to operation on the majority of standard Apostrophe 2 codebases. Projects that generally follow patterns in official documentation and sample projects will have the best results. The tool will help on very custom projects (especially the linting mode), but more manual work will be needed to finish the work.
```
boutell, line 128: "More manual work will always be needed" might read better and it's the truth

docs/guide/migration/upgrading.md
```diff
@@ -0,0 +1,134 @@
...
+- [Widget player code](/guide/custom-widgets.md#client-side-javascript-for-widgets) (client-side JavaScript) is not changed. The original widget player syntax relies on jQuery, which A3 does not include. A2 "lean" players may be supported in the future.
```
boutell, line 129: I would say:

A2 "lean" players may be automatically migrated in the future, however if you review the A3 widget player documentation you'll find it is very similar and easy to convert yourself.

docs/guide/migration/upgrading.md
boutell: Please include a mention that when upgrading it is best to replace all macros with fragments, and reemphasize that when you discuss the new area template tag.

boutell (changes requested):

docs/README.md
abea: redirects are in the enhanceApps.js file.

docs/guide/migration/upgrading.md
abea: On the latter point, that's why I specified "for each published document." I'm trying not to overcomplicate these descriptions.

docs/guide/migration/upgrading.md
abea: I'll update the example to changing the field name instead. Less radical, still breaks complete compatibility.

docs/guide/migration/upgrading.md
abea: I hesitate to start inserting comments on what is easy to update or not. If we do that for this and don't do it for other things it makes them seem harder. Or at least the easy stuff to update could seem harder.

docs/guide/migration/upgrading.md
boutell: 👍

boutell (approved):

docs/.vuepress/config.js
```diff
@@ -72,6 +72,10 @@ module.exports = {
           {
             text: 'Cookbook',
             link: '/cookbook/'
+          },
+          {
+            text: 'Migration',
+            link: '/guide/upgrading/'
```
agilbert, line 78: 404 here: https://a3.docs.apos.dev/guide/upgrading/

agilbert: Also it would be good to organize this list so it matches the order of the list on the docs index page.

agilbert (changes requested):
A few minor changes related to navigation.

agilbert (approved):

boutell (approved):

PR #156, opened 2021-12-01
PRO-2222 Documents l10n interpolation

abea:


docs/guide/localization/static.md
```diff
@@ -103,6 +104,53 @@ If you are customizing Apostrophe **user interface components**, the `$t()` meth
...
+
+String interpolation is a process of generating a text string that is partly dynamic. In the previous paragraph's example, the number of articles is dynamic, or variable, based on other input (the chosen filters).
+
+Regardless of whether we are localizing text in templates, server-side code, or UI Vue files, interpolation works essentially the same way. The string or localization key is still the first argument in the localization function. **Then an object is passed as a second argument to the l10n function, containing interpolation properties that match keys in curly braces.**
+
+Template example:
+
+<AposCodeBlock>
+  ```django
+    {{ __t('Displaying {{ count }} articles', {
```
boutell, line 119: If we're going to use `count` we really should cover pluralization too. We're supporting i18next JSON version 3, so the syntax is that if a second key is available in the JSON file with a `_plural` suffix it is used when count is not equal to `1`, otherwise the exactly matching key is used. This applies **only** if the interpolated value is named `count`.

boutell (changes requested):

docs/guide/localization/static.md
abea: That's valuable, but I don't want to complicate this section. I'll change the example.

boutell (approved):

PR #160, opened 2021-12-09
PRO-960 queries

abea:
This PR will be in a draft until the database query guide is also available.

### DIFF
--- a/docs/.vuepress/sidebar.js
+++ b/docs/.vuepress/sidebar.js
@@ -69,12 +69,13 @@ module.exports = {
           ]
         },
         [ 'guide/server-events.md', 'Server-side events' ],
+        [ 'guide/database-queries.md', 'Querying the database 🆕' ],
         [ 'guide/headless-cms.md', 'Using as a headless CMS' ],
         [ 'guide/hosting.md', 'Hosting in production' ],
         {
-          title: 'Other module config 🆕',
+          title: 'Other module config',
           children: [
-            [ 'guide/batch-operations.md', 'Batch operations 🆕' ]
+            [ 'guide/batch-operations.md', 'Batch operations' ]
           ]
         }
       ]
--- a/docs/.vuepress/sidebar.js
+++ b/docs/.vuepress/sidebar.js
@@ -116,13 +116,15 @@ module.exports = {
       ]
     },
     {
-      title: 'Module API',
+      title: 'Server-side',
       path: '/reference/module-api',
       collapsable: false,
       sidebarDepth: 2,
       children: [
         'reference/module-api/module-overview',
-        'reference/module-api/module-options'
+        [ 'reference/module-api/module-options', 'Module options' ],
+        '/reference/server-events',
+        '/reference/query-builders'
       ]
     },
     {
@@ -153,7 +155,6 @@ module.exports = {
       ]
     },
     'reference/template-tags',
-    'reference/server-events',
     {
       title: 'Core Modules',
       children: [
--- a/docs/reference/README.md
+++ b/docs/reference/README.md
@@ -1,7 +1,3 @@
 # Reference
 
-- [Glossary](/reference/glossary.md): Terminology and concepts that are important or useful to know when working with ApostropheCMS
-- [Field Types](/reference/field-types/): Core schema field types and their options
-- [REST API and additional API routes](/reference/api/): Built-in API routes for working with Apostrophe site content
-- [Module API](/reference/module-api/): Reference material for building and configuring Apostrophe modules
-- [Template tags](/reference/template-tags): Apostrophe-specific template tags that add additional functionality in templates.
+<AposRedirect path="/reference/glossary.html" />
--- a/docs/reference/query-builders.md
+++ b/docs/reference/query-builders.md
@@ -108,11 +108,11 @@ It is distinct from the `sort` builder to distinguish between cases where a defa
 
 In most custom uses the `sort` builder will likely be a more common option.
 
-<!--
 ### `distinctCounts()`
-TODO
+
 ```
-query.distinctCounts(true)
+const names = query.toDistinct('firstName')
+const counts = query.get('distinctCounts')
 ```
 
 **Default value:** `false`
@@ -120,7 +120,6 @@ query.distinctCounts(true)
 `.distinctCounts(true)` makes it possible to obtain counts for each distinct value after a call to `toCount()` is resolved by calling `query.get('distinctCounts')`. These are returned as an object in which the keys are the distinct values of the field, and the values are the number of occurrences for each value.
 
 This has a performance impact.
--->
 
 ### `_ids()`
 
--- a/docs/reference/query-builders.md
+++ b/docs/reference/query-builders.md
@@ -4,7 +4,7 @@ sidebarDepth: 2
 
 # Query builders
 
-Server-side database query builders in Apostrophe help developer refine database document queries or refine the query results in some way. All official query builders are written to either work as intended or to have no effect, at worst. The sections below organize core query builders by the relevant document types.
+Server-side database query builders in Apostrophe help developer refine database document queries or refine the query results in some way. The sections below organize core query builders by the relevant document types.
 
 **See the [database querying guide](#TODO) for general information about using query builders.**
 
@@ -237,7 +237,7 @@ query.project({
 
 The `project` builders sets the [MongoDB projection](https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/), specifying the document properties to included in the returned documents. The argument should be an object with properties of desired field names set to `1` to include those fields. Negative projections (`category: 0`) are currently not supported.
 
-In addition to database properties, Apostrophe provides two dynamic properties that can be included in hte `project` builder object:
+In addition to database properties, Apostrophe provides two dynamic properties that can be included in the `project` builder object:
 - `_url: 1` will include all properties required to generate an accurate `_url` property for each returned document.
 - `_relationshipName: 1` will add the properties required to permit a field of type `relationship` to be loaded, even though the related documents are not stored redundantly in the database. In this example, the name of the field is `_relationshipName`. `_articles: 1` could do the same for a relationship to `article` pieces.
 
--- a/docs/reference/query-builders.md
+++ b/docs/reference/query-builders.md
@@ -4,26 +4,13 @@ sidebarDepth: 2
 
 # Query builders
 
-Database query builders in Apostrophe help developer refine database document queries or refine the query results in some way. All official query builders are written to either work as intended or to have no effect, at worst. The sections below organize core query builders by the relevant document types.
+Server-side database query builders in Apostrophe help developer refine database document queries or refine the query results in some way. All official query builders are written to either work as intended or to have no effect, at worst. The sections below organize core query builders by the relevant document types.
 
 **See the [database querying guide](#TODO) for general information about using query builders.**
 
 ## Builders for all doc types
 
 The following query builders apply to all Apostrophe [doc types](/reference/glossary.md#doc).
-
-### `addLateCriteria()`
-
-```
-query.addLateCriteria({
-  geo: {
-    $near: [ 43.019590, -87.909290 ],
-    $maxDistance: 10
-  }
- })
-```
-
-The `addLateCriteria` builder is an opportunity to add query criteria to merge directly into the final criteria object that will go to MongoDB. This is meant to be used only in cases where MongoDB forbids the use of an operator inside an `$and` array, such as the `$near` operator.
 
 ### `addUrls()`
 
--- a/docs/reference/query-builders.md
+++ b/docs/reference/query-builders.md
@@ -108,19 +108,6 @@ It is distinct from the `sort` builder to distinguish between cases where a defa
 
 In most custom uses the `sort` builder will likely be a more common option.
 
-### `distinctCounts()`
-
-```
-const names = query.toDistinct('firstName')
-const counts = query.get('distinctCounts')
-```
-
-**Default value:** `false`
-
-`.distinctCounts(true)` makes it possible to obtain counts for each distinct value after a call to `toCount()` is resolved by calling `query.get('distinctCounts')`. These are returned as an object in which the keys are the distinct values of the field, and the values are the number of occurrences for each value.
-
-This has a performance impact.
-
 ### `_ids()`
 
 ```
--- a/docs/reference/query-builders.md
+++ b/docs/reference/query-builders.md
@@ -86,32 +86,6 @@ The `autocomplete` builder operates as a sort of fuzzy search for documents. It 
 
 This will only find partial matches in high-priority properties such as the `title`, or any string, select or checkbox field in the schema of the document.
 
-### `choices()`
-
-```
-query.choices('color, species')
-
-query.choices(['color', 'species'])
-```
-
-**Default value:** `false`
-
-The `choices` builder is use to populate a `choices` object on the query results based on the filters passed to it. Filters can be passed in as a single string of comma-separated values or an array of strings.
-
-That `choices` object will contain valid options for each filter based on the rest of the query. Those options can be used to populate a select field for filtering an index page, for example.
-
-### `counts()`
-
-```
-query.choices('color, species')
-
-query.choices(['color', 'species'])
-```
-
-**Default value:** `false`
-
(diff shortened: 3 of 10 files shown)


docs/.vuepress/sidebar.js
```diff
@@ -116,13 +116,15 @@ module.exports = {
       ]
     },
     {
-      title: 'Module API',
+      title: 'Server-side',
```
abea, line 119: It seemed like a good idea to start grouping these in a bigger bucket.

docs/reference/README.md
```diff
@@ -1,7 +1,3 @@
 # Reference
 
-- [Glossary](/reference/glossary.md): Terminology and concepts that are important or useful to know when working with ApostropheCMS
```
abea, line 3: We don't have a good plan to maintain this. I think we could create a proper landing page when the reference is more mature and we have a more complete structure.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+### `choices()`
+
+```
+query.choices('color, species')
+
+query.choices(['color', 'species'])
+```
+
+The `choices` builder is use to populate a `choices` object on the query results based on the filters passed to it. Filters can be passed in as a single string of comma-separated values or an array of strings.
```
boutell, line 102: This is incorrect re: how to get the results.

After awaiting `query.toArray`, you can invoke `query.get('choicesResults')` to get an object in which each property is the name of a field you passed to `choices` and each value is an array of choice objects with `label` and `value` properties.

This may seem a bit awkward but for most server side applications you would use `toChoices` or just `toDistinct`. This query builder was designed to help the `getAll` REST API populate its `choices` property.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+### `counts()`
+
+```
+query.choices('color, species')
+
+query.choices(['color', 'species'])
+```
+
+The `counts` builder works very similarly to the [`choices` builder](#choices). When passed a string of comma-separated filter names (or an array of filter names), it will include a `counts` object including the individual choices. Each choice will include a `count` property set to the number of documents in the query results that match that choice.
```
boutell, line 114: After awaiting `query.toArray`, you can invoke `query.get('countsResults')` to get an object in which each property is the name of a field you passed to `choices` and each value is an array of choice objects with `label`, `value` and `count` properties.

boutell: Again, server side you'd probably use `toChoices` with a single field name and an options object like `{ counts: true }`

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+```
+
+The `defaultSort` changes the default value for the [`sort` query builder](#sort).  The argument is the same as for the `sort` query builder: an object like `{ updatedAt: 1 }`, including a field name and `1` for ascending or `-1` for descending. See the [MongoDB `sort` method documentation](https://docs.mongodb.com/manual/reference/method/cursor.sort/) for other options. `false` can be passed as an option as well to clear the default sort.
+
+It is distinct from the `sort` builder to distinguish between cases where a default sort should be ignored (for instance, the `search` query builder is present) and cases where a sort is explicitly requested by the user.
+
+In most custom uses the `sort` builder will likely be a more common option.
+
+<!--
+### `distinctCounts()`
```
boutell, line 137: An implementation detail of `toChoices`.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+### `_ids()`
+
+```
+query._ids([
+  'ckwxzxdtr000t4v3rivysoari:en:published',
+  'ckwxzxdvc001b4v3rhxvxqgg1:en:published'
+])
+```
+
+The `_ids` builder causes the query to return only the document with matching `_id` properties and to return them in that order, assuming the documents with the specified IDs exist. All documents are fetched in the same locale regardless of the locale suffix of the values. If no locale can be determined via query parameters, the locale is inferred from the first _id in the set.
```
boutell, line 157: "fetched in the locale and mode of the request regardless..."

docs/reference/query-builders.md
boutell: It's not true that you can pass a string, that only works when "launder" is called, which only happens in the REST API (more specifically when `applyBuildersSafely` is used to apply the builders). We don't impose the performance hit of launder calls in pure server side programming, which is what we're creating a reference for here...

Well, sort of. Actually, since this is the reference for builders and some builders have a `launder` function, we *should* be talking about that and documenting what it accepts. But we have to make clear it's only applicable when the builder in question is present in the query string of a REST API request, or when `applyBuildersSafely` is used in a custom API.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+The query builder can also be called with a string, which is treated as a single document ID.
+
+### `limit()`
+
+```
+query.limit(10)
+```
+
+The `limit` query builder accepts an integer and limits the number of document results. It returns the first matching documents up to the set limit, though it does take the [`skip` builder](#skip) into account.
```
boutell, line 167: How about "after taking the"

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+### `locale()`
+
+```
+query.locale('es:published')
+```
+
+A valid locale identifier passed to the `locale` builder will tell the query to return results only belonging to that locale. Documents that match the query and have *no locale* (because their doc type is not localized) are also included in the results. Specifically passing `false` (the default value) will use the locale and mode on the `req` request object.
```
boutell, line 175: "to that locale, including the mode (draft or published) after a `:` character."

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+The `._url` property will include a site prefix if applicable and is always better to use than the document's `slug` for a URL.
+
+### `permission()`
+
+```
+query.permission('edit')
+```
+
+The `permission` builder is used restrict returned documents based on the action name passed as an argument. Only documents on which the `req` object can take the named action are returned. For example, using `.permission('edit')` on the query will only return documents that the requesting user (via `req`) can edit.
```
boutell, line 220: is used to restrict

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+Valid action arguments include:
+- `'view'`: The `req` has permission to view the documents. This is the default.
+- `'edit'`: The `req` has permission to edit the documents.
+- `'publish'`: The `req` has permission to publish the documents.
+- `false`: Bypass any permission checks. This returns everything regardless of permission level. *Use this with caution.*
+
+In all cases, all of the returned docs are marked with `_edit: true` properties if the user associated with the request is allowed to edit the document. This is useful if you are fetching docs for viewing but also want to know which ones can be edited.
```
boutell, line 228: `_edit: true` and `_publish: true`

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+### `perPage()`
+
+```
+query.perPage(20)
+```
+
+Using the `perPage` builder returns documents in sets of the number passed as an argument. This helps return documents in managable numbers and paginate the results, using the [`page` query builder](#page) to get specific set of results. This is usually easier than using `skip` and `limit` directly.
```
boutell, line 236: To get specific -> to get a specific

boutell: After the query completes, e.g. with `await query.toArray()` for instance, `query.get('totalPages')` will return the total number of pages.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+```
+query.project({
+  title: 1,
+  category: 1,
+  updatedAt: 1
+})
+```
+
+The `project` builders sets the [MongoDB projection](https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/), specifying the document properties to included in the returned documents. The argument should be an object with properties of desired field names set to `1` to include those fields.
+
```
boutell, line 260: Apostrophe extends `project` in the following ways:

* `_url: 1` will add all properties required to generate an accurate `_url` property for each returned document.
* `_relationshipName: 1` will add the properties required to permit a field of type `relationship` to be loaded (in this case, the name of the field is `_relationshipName`), even though the related documents are not stored redundantly in the database.

boutell: Negative projections, e.g. `body: 0`, are currently not supported.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+<!-- TODO: point to the direct db find method for the alternate way to project results. -->
```
boutell, line 261: Not sure that's necessary.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+### `search()`
+
+```
+query.search('tree')
+```
+
+The `search` builder limits results to those matching the string passed as an argument. Search is implemented using MongoDB's `$text` operator and a full text index.
+
+If this query builder is set, the `sort` query builder will default to sorting by search quality. This is important because the worst of the full-text search matches will be of very poor quality.
```
boutell, line 305: `search` does not support partial matches. For that, consider `autocomplete`.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+### `type()`
+
+```
+query.type('product')
+```
+
+The `type` builder can be used to limit a query to only one particular doc type. Pass the doc type name (string) as an argument.
+
+Generally you don't want to call this method directly. It will be better to include the doc type name as the `type` criteria in the original `find()` method arguments.
```
boutell, line 315: If you are invoking the `find` method of a particular piece type, then this is invoked for you and there is no need to worry about documents of the wrong type being returned. It is primarily useful in implementing special case "virtual piece types" like Apostrophe's `submitted-draft` module, which displays pieces of many types by extending its `find` method to pass `null` to this builder.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+```
+query.ancestors(true)
+
+query.ancestors({ children: true })
+```
+
+Setting the `ancestors` builder to `true` retrieves the array of "ancestors" for each returned page and assigns them to the `_ancestors` property. Page ancestors are those that precede a given page in the page tree hierarchy. The home page is `_ancestors[0]`. A returned page is not included in its own `_ancestors` array.
+
+If the builder argument is an object, do all of the above, and also call the query builders present in the object *on the query that fetches the ancestors*. For example, you can pass `{ children: true }` to fetch the children of each ancestor as the `_children` property of each ancestor, or pass `{ children: { depth: 2 } }` to get two layers of child pages.
```
boutell, line 339: "If the builder's argument is an object, the builder does all of the above, and also calls..."

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+The `children` builder is used to include the "children" of returned pages in an array on a `_children` property. If `children(true)` is called, it will return all children of a given page. If the argument is an object, it may have a `depth` property to fetch a specific number of child layers. Any
+other properties are passed on *as builders for the query that fetches the children*.
+
+### `isPage()`
+
+```
+query.isPage(true)
+```
+
+Passing `true` to the `isPage` builder ensures that results will only include documents that are pages. It defaults to `true` when used.
```
boutell, line 360: It defaults to `true`. There's no "when used". It's just that the builder only exists in a query that came from `apos.page.find`.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+```
+query.withPublished(true)
+```
+
+If the `withPublished` builder is set to `true`, then each document in the results will include a `_publishedDoc` property. That property will be set to the published version of that document. This is really only useful when querying for draft documents.
+
+## Page document query builders
+
+The following query builders only relate to page documents.
```
boutell, line 327: They are available on queries created via `apos.page.find` and, where supported, when using the REST API for pages. They are not available on generic queries created via `apos.doc.find`.

docs/reference/README.md
```diff
@@ -1,7 +1,3 @@
...
-- [Field Types](/reference/field-types/): Core schema field types and their options
-- [REST API and additional API routes](/reference/api/): Built-in API routes for working with Apostrophe site content
-- [Module API](/reference/module-api/): Reference material for building and configuring Apostrophe modules
-- [Template tags](/reference/template-tags): Apostrophe-specific template tags that add additional functionality in templates.
\ No newline at end of file
+<AposRedirect path="/reference/glossary.html" />
```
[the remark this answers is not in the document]
abea: It redirects to the glossary page. I'm using it a few other places.

docs/reference/README.md
abea: I guess I wasn't clear in my note. the structure of the reference section is evolving. Rather than continually updating this page and adding relatively obvious descriptors I'm instead redirecting to the glossary page and letting the sidebar navigation do the work.

boutell: ✅

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+## Builders for all doc types
+
+The following query builders apply to all Apostrophe [doc types](/reference/glossary.md#doc).
+
+### `addLateCriteria()`
+
+```
+query.addLateCriteria({
+  geo: {
+    $near: [ 43.019590, -87.909290 ],
```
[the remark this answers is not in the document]
abea: Yeah, if we have no use case for it we should deprecate it an plan for removing in A4. I'll remove this from documentation for now.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+    $maxDistance: 10
+  }
+ })
+```
+
+The `addLateCriteria` builder is an opportunity to add query criteria to merge directly into the final criteria object that will go to MongoDB. This is meant to be used only in cases where MongoDB forbids the use of an operator inside an `$and` array, such as the `$near` operator.
+
+### `addUrls()`
```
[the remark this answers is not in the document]
abea: For those that actually have a default set I put it in the desciption, but I can make it more standardized.

docs/reference/query-builders.md
```diff
@@ -0,0 +1,382 @@
...
+
+Passing `true` to the `attachments` builder will annotate all attachment fields in the returned documents with URLs and other metadata (primarily for images). This uses the the `apos.attachment.all` method with the `annotate: true` option. The builder defaults to `false`.
+
+### `autocomplete()`
+
+```
+query.autocomplete('tree house')
+```
+
+The `autocomplete` builder operates as a sort of fuzzy search for documents. It accepts a string and uses document search text properties to find matches. The string can contain multiple words, but each word (or segment) needs to be a partial string to find non-exact results. So "tree" will find results with "treehouse," but "treehouses" will not find "treehouse" (it is not meant as an advanced search tool, but can help find similar results).
```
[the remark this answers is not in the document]
abea: This is pulling some real threads. We don't have `weight` documented for fields and I would want to do real testing before we put this in writing. I've logged some general search configuration questions elsewhere. I don't think we have tests for that to fall back on. For now I guess we have to just say it's limited to those particular field types.

docs/reference/query-builders.md
abea: It sounds like this is not worth documenting since it's especially complicated and there are better ways for most code to do the same thing. What do you think?

docs/reference/query-builders.md
abea: Similar situation to `choices`. This might be better not documented and we document `toChoices` in the upcoming guide.

docs/reference/query-builders.md
```diff
@@ -256,9 +235,11 @@ query.project({
 })
 ```
 
-The `project` builders sets the [MongoDB projection](https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/), specifying the document properties to included in the returned documents. The argument should be an object with properties of desired field names set to `1` to include those fields.
+The `project` builders sets the [MongoDB projection](https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/), specifying the document properties to included in the returned documents. The argument should be an object with properties of desired field names set to `1` to include those fields. Negative projections (`category: 0`) are currently not supported.
 
-<!-- TODO: point to the direct db find method for the alternate way to project results. -->
+In addition to database properties, Apostrophe provides two dynamic properties that can be included in hte `project` builder object:
```
boutell, line 240: hte -> the

boutell (changes requested):

boutell (approved):

PR #162, opened 2021-12-16
PRO-897 data query guide

abea:
This is focusing on querying the database. I considered including update and insert methods as well, but this is long enough and that wasn't the assignment. A separate page on those will serve readers better.

### DIFF
--- a/docs/guide/database-queries.md
+++ b/docs/guide/database-queries.md
@@ -1,0 +1,1 @@
+# Querying the database
--- a/docs/guide/database-queries.md
+++ b/docs/guide/database-queries.md
@@ -1,1 +1,181 @@
+---
+sidebarDepth: 2
+---
+
 # Querying the database
+
+When designing custom modules or customizing existing modules, it can quickly become necessary to fetch data that Apostrophe does not automatically make available. We might want to display a random image in a page banner or a few dynamically-chosen articles related to the page a visitor is on. Apostrophe has an API designed to help developers get as creative as they need to be.
+
+One goal of the Apostrophe querying API is to facilitate common and advance requests to the MongoDB database *without* requiring developers to know advanced MongoDB syntax. Understanding that low-level syntax will be helpful at times, but advanced knowledge should not be necessary.
+
+As a final note, this guide will focus on working with content documents (e.g., pages and pieces). Querying database collections for things like the cache or long-running jobs will be covered elsewhere.
+
+::: note
+We use the terms "query" and "query builders" here. For developers with advanced Apostrophe 2 experience, these are generally the same as the A2 concepts of "cursors" and "cursor filters."
+:::
+
+## `find()`-ing data
+
+The page module (`@apostrophecms/page`) and all modules that *extend* the piece type module (`@apostrophecms/piece-type`), the two big "doc type" categories, have access to **a `find()` method** that initializes a data query. Any query using a doc type module's `find` method will be limited to that module type. This means that we can know that calling `self.find()` in a `product` piece module will only return "products".
+
+The `find` method takes up to three arguments:
+
+| Argument | Required | Description |
+| -------- | -------- | ----------- |
+| `req` | TRUE | The associated request object. Using a provided `req` object is important for maintaining user role permissions. |
+| `criteria` | FALSE | A [MongoDB criteria object](https://docs.mongodb.com/manual/tutorial/query-documents/). It is often as simple as properties that match schema field names assigned to the desired value. |
+| `options` | FALSE | The options object is converted to matching [query builders](#using-query-builders). It is often easier to add query builders in the method-syntax described below. |
+
+In the following example, we are writing the function that powers an [async template component](/guide/async-components.md). This hypothentical component would be passed the product `_id` property where it is shown as `productId`.
+
+<AposCodeBlock>
+  ```javascript
+  module.exports = {
+    extend: '@apostrophecms/piece-type',
+    // ...
+    components(self) {
+      return {
+        // Returning the five most recently created products.
+        async latest(req, data) {
+          // 👇 Setting up our query criteria.
+          const criteria = {
+            _id: { $ne: data.productId }
+          };
+
+          // 👇 The `find` method starting the query.
+          const products = await self.find(req, criteria)
+            .sort({ createdAt: -1 })
+            .limit(5)
+            .toArray();
+
+          return { products };
+        }
+      };
+    }
+  };
+  ```
+  <template v-slot:caption>
+    modules/product/index.js
+  </template>
+</AposCodeBlock>
+
+**What is happening here?**
+
+```javascript
+async latest(req, data) {
+  ...
+}
+```
+
+Component functions are provided a request object, `req`, usually from a website visitor visiting a page that uses the component. The function also receives `data`, which is a data object a developer included when using the component in a template. (See the [async component guide](/guide/async-components.md)] for more on that feature.)
+
+```javascript
+const criteria = {
+  _id: { $ne: data.productId }
+};
+```
+
+We are going to get the most recently added products, but it wouldn't help visitors to see the product they're already looking at in this group. Here we create a `criteria` object that uses the MongoDB query syntax to look for documents whose `_id` property is not equal to (`$ne`) the ID of the active product.
+
+```javascript
+await self.find(req, criteria)
+```
+
+We are in the `product` module, so by calling the `find` method from `self` (which is the `product` module itself), our results are automatically limited to products. That's why our `criteria` object does not need to include that limitation, which would otherwise look like `type: 'product'`.
+
+We pass in the `req` object we received from the component function parameters and the criteria object. In this case we are not including a third object for `options`.
+
+::: note
+It's important to understand that the `find` method is not the end of the process. It is only the beginning, setting up a query that may be refined and then needs to be executed by a separate method. There is more on that below, but for now make sure to understand that simply running `self.find()` by itself will not return any documents.
+:::
+
+```javascript
+  .sort({ createdAt: -1 })
+  .limit(5)
+```
+
+These are **query builders**. Let's look at what query builders are before we identify what is happening in this example.
+
+### Using query builders
+
+Query builders are additional instructions added to the data query. These special methods may receive arguments, but many apply an effect without any arguments. They can be chained on a query, meaning that we can add multiple builders onto a query and each one will build on the rules established before it.
+
+Examples of query builders include:
+- `.limit(10)`: This limits the number of results that the query will return to the number we pass it.
+- `.sort({ name: 1 })`: This builder sorts query results based on a document property value. This one sorts by the `name` property in ascending order.
+- `.search('search term')`: This accepts a string and looks for documents that have that string in their registered searchable words.
+- `.relationships(false)`: This builder accepts `false` or an array of relationship field names to avoid or limit relationship data loading onto the request response (saving some processing when it's not needed).
+
+As we can see here, query builders apply a wide variety of effects. See the [query builder reference page](/reference/query-builders.md) for a full list of query builders and how to use them.
+
+Let's take another look at the builders in the example above.
+
+```javascript
+self.find(req, criteria)
+  .sort({ createdAt: -1 })
+  .limit(5)
+```
+
+This example is using two query builders, `sort` and `limit`. As described above, `sort({ createdAt: -1})` sorts the queried documents in descending order on their `createdAt` properties (most recent first). `limit(5)` then ensures the request will return no more than five documents.
+
+::: note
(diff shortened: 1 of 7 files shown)


docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+
+```javascript
+query
+  .perPage(20)
+  .page(2)
+```
+
+This example tells the query that its results should be returned in a group of no more than 20 (`.perPage(20)`) and that we want the second group of results based on the active sort order (`.page(2)`). These builders are used in the Apostrophe REST APIs and can be very helpful when dealing with large amounts of content. They use two other builders internally (`limit` and `skip`), but can be easier to use for common situations.
+
+### Adding your own query builders
```
boutell, line 156: Before you say this, you should say that a query builder is automatically added for most schema field types (if it has an `addQueryBuilder` method in its field type definition, or extends one that does, then you can automatically use it as a query builder). This is a big part of what we mean when we say we save people effort over mastering MongoDB.

For relationship fields, there are several query builders added. If the field name is `_products`, then:

* `_products(product._id)` matches only documents related to the product with the given `_id`. You can also pass an array of product ids, in which case only documents related to *at least one of* those products will match.
* `products(product.slug)` (note no `_`) matches documents related to the product with that slug. You can also pass an array of product slugs, in which case only documents related to *at least one of* those products will match.
* `_productsAnd([ id, id, id...])` matches only documents related to *all* of the specified product ids (just one is not enough).
* `productsAnd([ slug, slug, slug...])` (note no `_`) matches only documents related to *all* of the specified product slugs (just one is not enough).

The main advantage of the query builders based on `_id` is that the `_id` can never change. The main advantage of the query builders based on slugs is that they are more user-friendly if they appear in the query string fo the URL.

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+Our query example from above uses `toArray()`. Let's look at that again.
+
+```javascript
+const products = await self.find(req, criteria)
+  .sort({ createdAt: -1 })
+  .limit(5)
+  .toArray();
+```
+
+As we've covered already, this code sets up the initial query with criteria (`self.find`) and adds builders with additional instructions (`sort` and `limit`). The final thing it does is run `toArray()` on the query. This is telling the database, "Take all the instructions we provided and give us back an array of results based on those instructions. Please." So, at the end of this, `products` will be an array of up to five document objects.
```
boutell, line 178: Take does not need to be capitalized.

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+
+::: note
+See that in our example we use `await` before `self.find`, indicating that this is running an asynchronous operation. It is worth noting that `toArray()` is the only asynchronous part of the code since that's the part that actually talks to the database.
```
boutell, line 181: 💯

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+```javascript
+const productId = 'ckcqi1ye1005mof3rdgtlln0b:en:published';
+
+const product = await self.find(req, { _id: productId })
+  .toObject();
+```
+
+### `toCount`
+
+The `toCount` query method is the easy and quickest way to simply get the number of documents that match a query. The `toCount` query method will ignore any `page`, `skip` and `limit` query builders in order to get the total number.
```
boutell, line 207: easiest

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+
+### `toChoices`
+
+The `toChoices` query method builds on `toDistinct` by returning each choice as an object with `label` and `value` properties. This can be useful when populating a select or other input field with options for a doc type property. For example, this is used for the document manager modal filter UI in Apostrophe.
+
+`toChoices` accepts an options object as an optional second argument. Setting `counts: true` in that options object will include a `count` property on each returned choice indicating how many documents match that choice.
+
+```javascript
+const teamOffices = await self.find(req, { type: 'team' })
+  .toDistinct('office', { count: true });
```
boutell, line 233: Should be `toChoices`, yeah?

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Using another module's `find` method
+
+The `self` object available in any Apostrophe module's customization functions can access other module managers on the `self.apos.modules` object. For example, if I'm working in the `article` module and want to query `author` pieces, I can access the `author` module with this:
+
+```javascript
+self.apos.modules.author
+```
+
+`self.apos.modules.author` is essentially the same as `self` would be if we were operating within the `author` module. Therefore we can query author pieces directly by using that module's `find` method.
```
boutell, line 248: It is exactly the same object.

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+  .toDistinct('office', { count: true });
+```
+
+## Query across modules
+
+The `find()` method in doc type modules is easy to use within each module as `self.find`. However this assumes we're looking for only content that is governed by the module where the method is called. There are two main ways to write these database queries from one module and get content managed by a separate module.
+
+### Using another module's `find` method
+
+The `self` object available in any Apostrophe module's customization functions can access other module managers on the `self.apos.modules` object. For example, if I'm working in the `article` module and want to query `author` pieces, I can access the `author` module with this:
```
boutell, line 242: Not sure we need to talk about managers here, but if we do, then it's "other doc type manager modules," not the other way around.

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+`self.apos.modules.author` is essentially the same as `self` would be if we were operating within the `author` module. Therefore we can query author pieces directly by using that module's `find` method.
+
+```javascript
+const activeAuthors = self.apos.modules.author.find(req, { active: true })
+  .toArray();
+```
+
+If a doc type module has been [assigned an alias](/reference/module-api/module-options.md#alias), the doc type's manager will be directly on `self.apos`. For example, the `@apostrophecms/page` module is available as `self.apos.page` because `'page'` is its alias option value.
```
boutell, line 255: How about "then that doc type module will be..."

Manager is different language than the beginning of the sentence and probably doesn't need to be brought up now that the module name and the doc `type` in the database are always equal.

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+There is a module that governs all content documents: `@apostrophecms/doc`, which is aliased as `'doc'`. Its `find` method works the same as any individual doc type's, but it is not restricted to any one document type. We can then use it to look for a particular type with the `type` property or pass it criteria and get documents of multiple types in the results.
+
+```javascript
+const featuredContent = self.apos.doc.find(req, { featured: true })
+  .toArray();
+
+
+const featuredByType = self.apos.doc.find(req, {
+  featured: true,
+  type: selectedType // A hypothetical option a user selected.
```
boutell, line 270: We should mention the limitations here. Pieces won't get their `_url` property this way, so you might have to make an additional query with each type-specific `find` after doing your blanket query across types, perhaps to get them all sorted by search match quality... which is really the job of the search module, and almost but not quite the only use case for this that most developers will encounter.

boutell (changes requested):

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
+# Querying the database
+
+When designing custom modules or customizing existing modules, it can quickly become necessary to fetch data that Apostrophe does not automatically make available. We might want to display a random image in a page banner or a few dynamically-chosen articles related to the page a visitor is on. Apostrophe has an API designed to help developers get as creative as they need to be.
+
+One goal of the Apostrophe querying API is to facilitate common and advance requests to the MongoDB database *without* requiring developers to know advanced MongoDB syntax. Understanding that low-level syntax will be helpful at times, but advanced knowledge should not be necessary.
```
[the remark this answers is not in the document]
abea: > just to use the word one more time

I'm not sure what you're suggesting there.

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+
+As a final note, this guide will focus on working with content documents (e.g., pages and pieces). Querying database collections for things like the cache or long-running jobs will be covered elsewhere.
```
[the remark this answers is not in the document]
abea: good point. I'll update the earlier API reference to "Apostrophe content query" to allude to this without complicating the idea further.

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+| -------- | -------- | ----------- |
+| `req` | TRUE | The associated request object. Using a provided `req` object is important for maintaining user role permissions. |
+| `criteria` | FALSE | A [MongoDB criteria object](https://docs.mongodb.com/manual/tutorial/query-documents/). It is often as simple as properties that match schema field names assigned to the desired value. |
+| `options` | FALSE | The options object is converted to matching [query builders](#using-query-builders). It is often easier to add query builders in the method-syntax described below. |
+
+In the following example, we are writing the function that powers an [async template component](/guide/async-components.md). This hypothentical component would be passed the product `_id` property where it is shown as `productId`.
+
+<AposCodeBlock>
+  ```javascript
+  module.exports = {
```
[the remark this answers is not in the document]
abea: the file name is below in the `<template v-slot:caption>` slot

docs/guide/database-queries.md
```diff
@@ -0,0 +1,273 @@
...
+```javascript
+await self.find(req, criteria)
+```
+
+We are in the `product` module, so by calling the `find` method from `self` (which is the `product` module itself), our results are automatically limited to products. That's why our `criteria` object does not need to include that limitation, which would otherwise look like `type: 'product'`.
+
+We pass in the `req` object we received from the component function parameters and the criteria object. In this case we are not including a third object for `options`.
+
+::: note
+It's important to understand that the `find` method is not the end of the process. It is only the beginning, setting up a query that may be refined and then needs to be executed by a separate method. There is more on that below, but for now make sure to understand that simply running `self.find()` by itself will not return any documents.
```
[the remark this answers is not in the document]
abea: I've been there.

docs/guide/database-queries.md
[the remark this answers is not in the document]
boutell: Just to say "query API" rather than "querying API". But not a big deal.

docs/guide/database-queries.md
```diff
@@ -128,12 +143,13 @@ One very useful query builder is `.project()`. We often know what specific data
...
   .project({
     title: 1,
+    _url: 1,
     _author: 1,
     publishedAt: 1
   })
 ```
 
-The `project` builder is one that uses a MongoDB syntax. The object passed as an argument should include schema field names for the document type we're querying with each set to `1`. This tells the database that we only want these fields in each document we get back.
+The `project` builder is one that uses a MongoDB syntax. The object passed as an argument should include schema field names for the document type we're querying with each set to `1`. This tells the database that we only want these fields in each document we get back. In addition to simple fields, Apostrophe enhances `project` to handle relationship fields and the special property `_url`.
```
boutell, line 152: Any truthy value will do but `1` is the convention, yeah.

docs/guide/database-queries.md
```diff
@@ -153,6 +169,65 @@ query
...
+- `relationship`
+
+`relationship` fields get special treatment with four query builders for each field. If the field name is `_products`, then:
+
+- `._products(product._id)` matches only documents related to the product with the provided `_id`. You can also pass an array of product IDs, in which case documents related to *at least one* of those products will match.
+- `.products(product.slug)` (no leading `_`) is similar, but supports matching by the `slug` property. You can also pass an array of product slugs, in which case only documents related to *at least one* of those products will match.
+- `._productsAnd([ _id1, _id2, _id3...])` matches only documents related to *all* of the specified product IDs (one relation is not enough).
+- `.productsAnd([ slug1, slug2, slug3...])` (no leading  `_`) matches only documents related to *all* of the specified product slugs (one relation is not enough).
+
+`_id` properties are useful since they will never change, but slugs can be more readable (for humans) and parsed from URLs.
```
boutell, line 229: I'd say "when visible in URLs," since both ids and slugs work fine in URLs — the issue is readability.

boutell (changes requested):
Extremely minor language thing. Then good to go.

docs/guide/database-queries.md
abea: I'll say `1` or `true` since that's what MongoDB docs list. Even if other truthy values will work I don't think it helps devs to carry any more than this in their head. https://docs.mongodb.com/manual/reference/method/db.collection.find/#projection

boutell (approved):

PR #168, opened 2022-01-10
PRO-2415 Drafts the advanced guide on inserting and updating docs

abea:
## Summary

Resolves PRO-2415 by drafting a guide on inserting and updating docs server-side.

n/a. Review the content for accuracy and clarity.

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Documentation
- [ ] Build-related changes
- [ ] Other

- [x] The changelog is updated
- [x] Related documentation has been updated
- [x] Related tests have been updated

**Other information:**

### DIFF
--- a/docs/guide/database-insert-update.md
+++ b/docs/guide/database-insert-update.md
@@ -82,7 +82,7 @@ return updatedDraft;
 
 We update the document property that tracks the dog's adoption status on the data object and use `self.update` to replace the previous document state with our update (with `await` as it is asynchronous). We finally return the result, which will be the updated document object.
 
-Note that using the provided `req` object like this works only if the `req` object is from a user with *at least* "contributor" permissions for the `dog` piece type. We could pass `{ permissions: false }` as a third options argument to `self.update()` to bypass that permission check, though that obviously raises security issues we would need to consider carefully.
+Note that using the provided `req` object like this works only if the `req` object is from a user with *at least* "contributor" permissions for the `dog` piece type. **If we wanted to bypass that permission check**, or if we ever wanted to allow anonymous site visitors to insert or update content, we would pass `{ permissions: false }` as a third options argument to `self.update()`. That obviously raises security issues we would need to consider carefully.
 
 ::: note
 All content documents have multiple versions, including "draft" and "published" versions. The `update()` methods only updates the "draft" copy, allowing editors to still review before publishing. If we *did* want to publish here as well, we would want to run the publishing method:
--- a/docs/guide/database-insert-update.md
+++ b/docs/guide/database-insert-update.md
@@ -75,14 +75,22 @@ If the identified dog isn't in our database yet, we would return a `null` value 
 // We update the `status` property.
 dogDocument.status = status;
 
-const updateResult = await self.update(req, dogDocument);
+const updatedDraft = await self.update(req, dogDocument);
 
-return updateResult;
+return updatedDraft;
 ```
 
-We update the document property that tracks the dog's adoption status on the data object and use `self.update` to replace the previous document state with our update (with `await` as it is asynchronous). We finally return the result, which will be an object with update information from MongoDb.
+We update the document property that tracks the dog's adoption status on the data object and use `self.update` to replace the previous document state with our update (with `await` as it is asynchronous). We finally return the result, which will be the updated document object.
+
+Note that using the provided `req` object like this works only if the `req` object is from a user with *at least* "contributor" permissions for the `dog` piece type. We could pass `{ permissions: false }` as a third options argument to `self.update()` to bypass that permission check, though that obviously raises security issues we would need to consider carefully.
 
 ::: note
+All content documents have multiple versions, including "draft" and "published" versions. The `update()` methods only updates the "draft" copy, allowing editors to still review before publishing. If we *did* want to publish here as well, we would want to run the publishing method:
+
+```javascript
+await self.publish(req, updatedDraft);
+```
+
 **What if we are updating a document from a different doc type module?**
 
 In the example above, `self` refers to the `dog` piece type module since that is where the method is registered. If we wanted to run the `update` method from a separate module we would replace `self` with a specific reference to the `dog` module: **`self.apos.modules.dog`**.
@@ -92,9 +100,9 @@ In the example above, `self` refers to the `dog` piece type module since that is
 
 Updating pages works the same way as pieces with the same arguments to the `update` method. There are a few things to keep in mind when working on pages, however.
 
-**We often call the page `update` method from `self.apos.page`.** `self.update` does work from individual page type modules and sometimes that is the right way to call it. But since pages can change their `type` property (unlike pieces) they share a single `update` method from the `@apostrophecms/page` module. `self.update` on individual page type modules is simply a wrapper around that method.
+**We typically call the page `update` method from the main page module: `self.apos.page.update()`.** Since pages can change their `type` property (unlike pieces) they share a single `update` method from the `@apostrophecms/page` module. `self.update()` on individual page type modules is simply a wrapper around that method.
 
-**There is a better way to move pages within the [page tree](/guide/pages.md#connecting-pages-with-page-tree-navigation).** The `@apostrophecms/page` module has a dedicated `move()` method for that purpose. Documentation of that method is upcoming. In the meantime, refer to the [REST API information about page tree placement](/reference/api/pages.md#post-api-v1-apostrophecms-page).
+**`update()` is not the way to move pages within the [page tree](/guide/pages.md#connecting-pages-with-page-tree-navigation).** The `@apostrophecms/page` module has a dedicated `move()` method for that purpose. Documentation of that method is upcoming. In the meantime, refer to the [REST API information about page tree placement](/reference/api/pages.md#post-api-v1-apostrophecms-page).
 
 ## Inserting a new piece
 
--- a/docs/guide/database-insert-update.md
+++ b/docs/guide/database-insert-update.md
@@ -1,0 +1,159 @@
+# Inserting and updating in server-side code
+
+Apostrophe provides module methods for inserting and updating content documents in the database. These methods should be the primary tools for developers doing server-side data operations since they [emit events](/guide/server-events.md) that the CMS uses to keep all database collections updated and in sync. We will look at methods for both pieces and pages, which are similar but have some important differences.
+
+::: note
+Remember that the Apostrophe REST APIs for [pieces](/reference/api/pieces.md) and [pages](/reference/api/pages.md) are usually the best option when triggering content updates from the browser. Those API endpoints take advantage of the methods below while adding logic important to use in browsers.
+:::
+
+## Updating content documents
+
+When we have an existing piece document that to update, we use the `update()` method on the related piece type module. The `update` method takes the following arguments:
+
+| Argument | Required | Description |
+| -------- | -------- | ----------- |
+| `req` | TRUE | The associated request object. Using a provided `req` object is important for maintaining user role permissions. |
+| `piece` | TRUE | The document object that will *replace* the existing database document. |
+| `options` | FALSE | An options object, currently only used for internal draft state management. |
+
+For example, if we had a dog adoption website we may want to update `dog` pieces to indicate whether someone requested information about them. This could be after a contact form receives a submission about the dog.
+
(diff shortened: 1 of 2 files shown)


docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+
+<AposCodeBlock>
+  ```javascript
+  module.exports = {
+    extend: '@apostrophecms/piece-type',
+    // ...
+    methods (self) {
+      // The hypothetical contact form can only pass along the `_id` for the dog.
+      async registerInterest(dogId) {
+        const req = apos.task.getReq({ role: 'editor' });
```
boutell, line 29: Feel like this immediately bypasses what was said earlier about the importance of using a provided req object. There's a reason (in this context the site visitor probably has no permissions), but that should be explained before we run right out and use this.

docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+        const dogDocument = await self.find(req, { _id: dogId }).toObject();
```
boutell, line 30: Result could be null.

docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+
+        // We update the `receivedInterest` integer property.
+        dogDocument.receivedInterest++;
```
boutell, line 33: This is actually a better use case for a direct mongodb `$inc` operation on the published version of the doc, to avoid race conditions when many people are expressing interest in a dog, or loss of work if someone is trying to edit the dog while people are expressing interest. It's similar to the "hit counter" use case, in other words. So maybe not the best example of this.

Using `insert` and `update` is most helpful when you want the permissions checks and you want the beforeSave/afterSave etc. stuff to happen.

docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+
+```javascript
+async registerInterest(dogId) {
+  const req = apos.task.getReq({ role: 'editor' });
+  const dogDocument = await self.find(req, { _id: dogId }).toObject();
+  ...
+}
+```
+
+Our method received the `_id` of the document associated with this dog. We then generate a `req` object using `apos.task.getReq()`. While it is best to use a "natural" request object, if the contact form submission was from someone not logged in then the `req` will not have permission to update documents. We would want to make sure to guard against spam submissions elsewhere.
```
boutell, line 57: OK I see that you did explain the reason for using a task req here.

docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+**What if we are updating a document from a different doc type module?**
+
+In the example above, `self` refers to the `dog` piece type module since that is where the method is registered. If we wanted to run the `update` method from a separate module we would replace `self` with a specific reference to the `dog` module: **`self.apos.modules.dog`**.
+:::
+
+### Updating page documents
+
+Updating pages works the same way as pieces with the same arguments to the `update` method. There are a few things to keep in mind when working on pages, however.
+
+**We often call the page `update` method from `self.apos.page`.** `self.update` does work from individual page type modules and sometimes that is the right way to call it. But since pages can change their `type` property (unlike pieces) they share a single `update` method from the `@apostrophecms/page` module. `self.update` on individual page type modules is simply a wrapper that method.
```
boutell, line 81: wrapper around that method.

docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+  ```javascript
+  module.exports = {
+    extend: '@apostrophecms/piece-type',
+    // ...
+    methods (self) {
+      async addNewDog(initialInfo) {
+        // Generate a req object using the contributor role
+        const req = apos.task.getReq({ role: 'contributor' });
+        // Generate a blank dog data object.
+        let newDog = self.newInstance();
```
boutell, line 101: This example works better than the update example because it always makes sense to call `self.insert` (or page.insert) when creating a new thing.

docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+
+The arguments for the `self.apos.page.insert()` method are:
+
+| Argument | Required | Description |
+| -------- | -------- | ----------- |
+| `req` | TRUE | The associated request object. Using a provided `req` object is important for maintaining user role permissions. |
+|`_targetId` | TRUE | The `_id` of an existing page to use as a target when inserting the new page. `_home` and `_archive` are optional conveniences for the home page and [archived section](/reference/api/pages.md#moving-pages-to-the-archive), respectively. |
+|`_position` | TRUE | A numeric value will represent the zero-based child index under the `_targetId` page. `before`, `after`, `firstChild`, or `lastChild` values set the position within the page tree for the new page in relation to the target page (see `_targetId`). `before` and `after` insert the new page as a sibling of the target. `firstChild` and `lastChild` insert the new page as a child of the target. |
```
boutell, line 130: With all this covered here, you're not that far from completing coverage of the `move` method too. Just sayin'.

docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+
+  await self.apos.page.insert(req, '_home', 'lastChild', newPage);
+  ```
+  <template v-slot:caption>
+    modules/special-page/index.js
+  </template>
+</AposCodeBlock>
+
+::: note
+Whenever updating or inserting a content document, do not attempt to directly set any property beginning with an underscore (`_`). Any property with a leading underscore, except for `_id`, **will not be saved to the database**.
```
boutell, line 154: While true in A2, this is incorrect in A3. In A3, relationship fields are read/write at the apostrophe model API level.

That means that if product has a relationship to article, you can populate `._articles` with an array of articles and insert your product and it'll save that relationship. However note that it won't save the articles themselves, you have to do that separately first if the articles themselves are new or updated. This also means it's OK to use article objects with a strict projection in the array, since you're not modifying them, just the relationship to them (apostrophe only really cares about the `_id` property and, if there is a relationship schema, the `_relationship` property of each one).

docs/guide/database-insert-update.md
```diff
@@ -0,0 +1,159 @@
...
+
+Apostrophe only allows starting property names with an underscore on dynamically loaded information. This includes [populated relationship information](/guide/relationships.md) and the `_url` property. These properties can change at any time and make use of other information stored elsewhere in the database.
+
+With very few exceptions, the `_id` property is an automatically generated, randomized, and (always) unique property. It is also special in that it can never change for a given document.
```
boutell, line 158: Correct. You should not try to set it yourself because you're inviting problems with Apostrophe's modes and locales if you do.

boutell (changes requested):

docs/guide/database-insert-update.md
abea: Right. Maybe there's a better example to use here.

boutell: Mmm. Something in a command line task would be good.

docs/guide/database-insert-update.md
abea: This is copied from the REST API pages. I'll be documenting `move` method in the page module's reference page and thought that'd be better referenced from there.

docs/guide/database-insert-update.md
abea: Ah, right. I'll remove this, then and leave the bit about _id.

docs/guide/database-insert-update.md
```diff
@@ -47,26 +53,34 @@ For example, if we had a dog adoption website we may want to update `dog` pieces
...
 
-We then request the *full* database document. See the [database querying guide](/guide/database-queries.md) for more on this.
+If the identified dog isn't in our database yet, we would return a `null` value and the update task's function could insert the dog instead.
 
 ```javascript
-dogDocument.receivedInterest++;
+// We update the `status` property.
+dogDocument.status = status;
 
 const updateResult = await self.update(req, dogDocument);
```
boutell, line 78: This does read better, but we should mention that:

* This will only work if `req` is for a user with permission to at least contribute to dogs. You can pass `{ permissions: false }` as the options argument to `update` if you want to allow the public or a guest user to update the status of a dog in this way.
* The update method will only modify the current draft, unless the piece type has the`localize: false` option or the `autopublish: true` option. There is a publish method available, if you want to follow up with that in a particular case. Also a submit method, if you want to treat it as a contribution so someone with permission can confirm it.

Just don't want people to turn right around and say it "doesn't work."

docs/guide/database-insert-update.md
```diff
@@ -47,26 +53,34 @@ For example, if we had a dog adoption website we may want to update `dog` pieces
...
 
 return updateResult;
 ```
 
-We then update the document property that tracks how many times someone expressed interested in the dog. Then we use `self.update` to replace the previous document state with our update (with `await` as it is asynchronous). We finally return the result, which will be an object with update information from MongoDb.
+We update the document property that tracks the dog's adoption status on the data object and use `self.update` to replace the previous document state with our update (with `await` as it is asynchronous). We finally return the result, which will be an object with update information from MongoDb.
```
boutell, line 83: The result of `self.update(req, doc, options)` will be the updated document itself, not a mongodb info object.

docs/guide/database-insert-update.md
```diff
@@ -78,7 +92,7 @@ In the example above, `self` refers to the `dog` piece type module since that is
 
 Updating pages works the same way as pieces with the same arguments to the `update` method. There are a few things to keep in mind when working on pages, however.
 
-**We often call the page `update` method from `self.apos.page`.** `self.update` does work from individual page type modules and sometimes that is the right way to call it. But since pages can change their `type` property (unlike pieces) they share a single `update` method from the `@apostrophecms/page` module. `self.update` on individual page type modules is simply a wrapper that method.
+**We often call the page `update` method from `self.apos.page`.** `self.update` does work from individual page type modules and sometimes that is the right way to call it. But since pages can change their `type` property (unlike pieces) they share a single `update` method from the `@apostrophecms/page` module. `self.update` on individual page type modules is simply a wrapper around that method.
```
boutell, line 95: I would just skip the thing about calling it from individual page type modules instead. It's never "wrong" to call it from the page module and it's potentially wrong not to.

docs/guide/database-insert-update.md
```diff
@@ -78,7 +92,7 @@ In the example above, `self` refers to the `dog` piece type module since that is
...
 
 **There is a better way to move pages within the [page tree](/guide/pages.md#connecting-pages-with-page-tree-navigation).** The `@apostrophecms/page` module has a dedicated `move()` method for that purpose. Documentation of that method is upcoming. In the meantime, refer to the [REST API information about page tree placement](/reference/api/pages.md#post-api-v1-apostrophecms-page).
```
boutell, line 97: It's not a better way, it's the only way at the JS api level (update can't do it).

boutell (changes requested):

docs/guide/database-insert-update.md
abea: I was thinking that it was possible with a mess of extra steps, but it's an unnecessary mystery.  👍

docs/guide/database-insert-update.md
```diff
@@ -16,21 +16,27 @@ When we have an existing piece document that to update, we use the `update()` me
 | `piece` | TRUE | The document object that will *replace* the existing database document. |
 | `options` | FALSE | An options object, currently only used for internal draft state management. |
 
-For example, if we had a dog adoption website we may want to update `dog` pieces to indicate whether someone requested information about them. This could be after a contact form receives a submission about the dog.
+For example, if we had a dog adoption website that used an external API, we may want to update `dog` pieces periodically with adoption status. This may be using a command line task that is run regularly by a cron job.
```
boutell, line 19: OK, so `req` is going to be a task req with admin permissions in your example, but I feel like we're just asking for a lot of frustrated devs if we don't mention the `{ permissions: false }` option is the right way to go if you *do* want to let an anonymous site visitor insert or update a piece.

boutell (changes requested):

docs/guide/database-insert-update.md
abea: I updated the paragraph about this to try to highlight it a bit more.

boutell (approved):

PR #176, opened 2022-01-18
PRO-2443 documented nestedModuleSubdirs for a3

boutell:
## Summary

Documents PRO-2442, nestedModuleSubdirs support for A3.

Review the docs.

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [X] Documentation
- [ ] Build-related changes
- [ ] Other

- [N/A] The changelog is updated
- [X] Related documentation has been updated
- [N/A] Related tests have been updated

**Other information:**

### DIFF
--- a/docs/guide/nested-module-subdirs.md
+++ b/docs/guide/nested-module-subdirs.md
@@ -1,0 +1,113 @@
+# Nested module folders
+
+Apostrophe has optional support for nested subdirectories of modules, tucked
+inside `modules`.
+
+You must set the `nestedModuleSubdirs` option to `true` in `app.js`, like this:
+
+
+<AposCodeBlock>
+```javascript
+require('apostrophe')({
+  shortName: 'my-project',
+  nestedModuleSubdirs: true
+  // etc., you may have additional options here as always
+  modules: {
+    // You can still enable all modules here, or move some or all to modules.js
+    // files in subdirectories, as seen below
+  }
+});
+```
+  <template v-slot:caption>
+    app.js
+  </template>
+</AposCodeBlock>
+
+Now you can nest modules in subdirectories, like the example `modules/products` folder explored below. Start with a `modules.js` file in the parent `modules/products` folder. Here you can activate all of the modules that relate to products, making `app.js` shorter:
+
+<AposCodeBlock>
+```javascript
+// modules/products/modules.js
+module.exports = {
+  'product': {},
+  'product-page': {},
+  'product-widget': {}
+};
+```
+  <template v-slot:caption>
+    modules/products/modules.js
+  </template>
+</AposCodeBlock>
+
+And then you can implement those modules in their own sub-subdirectories:
+
+<AposCodeBlock>
+```javascript
+// modules/products/product/index.js
+module.exports = {
+  extend: '@apostrophecms/piece-type',
+  options: {
+    label: 'Products'
+  }
+};
+```
+  <template v-slot:caption>
+    modules/products/product/index.js
+  </template>
+</AposCodeBlock>
+
+<AposCodeBlock>
+```javascript
+// modules/products/product-page/index.js
+module.exports = {
+  extend: '@apostrophecms/piece-page-type',
+  options: {
+    label: 'Products Page'
+  }
+};
+```
+  <template v-slot:caption>
+    modules/products/product-page/index.js
+  </template>
+</AposCodeBlock>
+
+<AposCodeBlock>
+```javascript
+// modules/products/product-widget/index.js
+module.exports = {
+  extend: '@apostrophecms/widget-type',
+  options: {
+    label: 'Products'
+  },
+  fields: {
+    add: {
+      _products: {
+        label: 'Products',
+        type: 'relationship',
+        withType: 'product'
+      }
+    }
+  }
+};
+```
+  <template v-slot:caption>
+    modules/products/product-widget/index.js
+  </template>
+</AposCodeBlock>
+
+The resulting directory tree looks like this:
+
+```
+/app.js
+/modules
+/modules/products (modules.js lives here, activates three modules)
+/modules/products/product (index.js for the product piece type lives here)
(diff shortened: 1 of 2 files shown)


docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+<AposCodeBlock>
+```javascript
+require('apostrophe')({
+  shortName: 'my-project',
+  nestedModuleSubdirs: true
+  // etc., you may have additional options here as always
+  modules: {
+    // You can still enable all modules here, or move some or all to modules.js
```
abea, line 16: Let's introduce `'products': {}` here as well. We can call that out and say we're including that as the start of the example. This is all we need to get the others once they're registered.

docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
...
+  <template v-slot:caption>
+    modules/products/modules.js
+  </template>
+</AposCodeBlock>
+
+And then you can implement those modules in their own sub-subdirectories:
+
+<AposCodeBlock>
+```javascript
+// modules/products/product/index.js
```
abea, line 46: this line isn't needed anymore since it's in the caption.

docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
...
+};
+```
+  <template v-slot:caption>
+    modules/products/product/index.js
+  </template>
+</AposCodeBlock>
+
+<AposCodeBlock>
+```javascript
+// modules/products/product-page/index.js
```
abea, line 61: same here.

docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
...
+};
+```
+  <template v-slot:caption>
+    modules/products/product-page/index.js
+  </template>
+</AposCodeBlock>
+
+<AposCodeBlock>
+```javascript
+// modules/products/product-widget/index.js
```
abea, line 76: same

docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
...
+And then you can implement those modules in their own sub-subdirectories:
+
+<AposCodeBlock>
+```javascript
+// modules/products/product/index.js
+module.exports = {
+  extend: '@apostrophecms/piece-type',
+  options: {
+    label: 'Products'
+  }
```
abea, line 51: I suggest adding a line below: `// Other configuration...`

docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
...
+    // files in subdirectories, as seen below
+  }
+});
+```
+  <template v-slot:caption>
+    app.js
+  </template>
+</AposCodeBlock>
+
+Now you can nest modules in subdirectories, like the example `modules/products` folder explored below. Start with a `modules.js` file in the parent `modules/products` folder. Here you can activate all of the modules that relate to products, making `app.js` shorter:
```
abea, line 26: This would be the place to say that the `products` directory name is arbitrary. It could even be `sales-modules` to be extra distinct.

docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
...
+```
+/app.js
+/modules
+/modules/products (modules.js lives here, activates three modules)
+/modules/products/product (index.js for the product piece type lives here)
+/modules/products/product-page-type (index.js, views/show.html, etc.)
+/modules/products/product-widget (index.js, views/widget.html, etc.)
+```
+
+Just remember: **the names of the parent folders do not matter, and the names of the actual
```
abea, line 109: Let's make sure we mention this above as well. It could be good to emphasize that everything about those subdirectories will work the same as they would directly in `modules`.

docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
...
+module folders at the bottom MUST still match the name of each module.**
+
+By following through with this approach you can make `app.js` much shorter and better organize your project.
```
abea, line 112: Something about "making `app.js` shorter" seems to be underselling. More organized and less verbose seem like better talking points to lean on.

abea (changes requested):

docs/guide/nested-module-subdirs.md
```diff
@@ -0,0 +1,113 @@
+# Nested module folders
+
+Apostrophe has optional support for nested subdirectories of modules, tucked
```
abea, line 3: This introductory paragraph could be clearer about what the feature is. I'd suggest leading with a problem this solves to contextualize it. "As website projects grow, code organization is increasingly important. To help with this, Apostrophe support optionally nesting module directories and registering them as module clusters."

I threw in the clustering bit. Simply nesting is a codebase file benefit. Registering them as groups with a `modules.js` file in app.js is an additional code organization benefit, considering these modules as a group (or cluster) of some sort.

abea (changes requested):

docs/guide/nested-module-subdirs.md
boutell: I don't think I understand — there's no such feature. `modules.js` files are loaded automatically, otherwise the subdirs themselves have no significance other than convenience, that leaves no role for placing `modules: {}` here. Unless you're talking about `module: {}` (the piece type), which also doesn't make sense in isolation as near as I can tell.

abea: Okay, we've discussed this outside of this thread so I understand better.

The docs PR could definitely be clearer on the fact that `modules.js` is not necessary. In this example it seems there's no need for a modules.js file, so we should leave it out. If you want to include that feature we should include a clear case for when it would be used.

abea (changes requested):

abea (approved):

PR #180, opened 2022-01-24
PRO-2465 module refence page

abea:
## Summary

Documents the base module for Apostrophe

Review for accuracy and clarity.

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/docs/reference/modules/_template.md
+++ b/docs/reference/modules/_template.md
@@ -1,8 +1,14 @@
+---
+extends: '@apostrophecms/module'
+---
+
 # `@apostrophecms/module-name`
 
 **Alias:** `apos.i18n`
 
 *General description paragraph.* Cras mattis consectetur purus sit amet fermentum. Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+
+**Extends:** `{{ $frontmatter.extends }}`
 
 ## Options
 
@@ -27,7 +33,7 @@ The following methods belong to this module and may be useful in project-level c
 
 Because this module has an alias, you can call these from another module from the alias path. For example, `self.apos.[the alias].inferIdLocaleAndMode()`.
 
-### `inferIdLocaleAndMode(req, _id)`
+### `async inferIdLocaleAndMode(req, _id)`
 
 Curabitur blandit tempus porttitor. Nullam quis risus eget urna mollis ornare vel eu leo. Nullam id dolor id nibh ultricies vehicula ut id elit.
 
--- a/docs/reference/modules/module.md
+++ b/docs/reference/modules/module.md
@@ -1,0 +1,103 @@
+# `@apostrophecms/module`
+
+This module is the foundation of all other Apostrophe modules. It is not *directly* used for any functionality, but [every other module inherits its behavior and functionality](/guide/modules.md#module-inheritance). Every other module has access to use the features documented here. Critically, this module includes initialization behavior that all modules need to start up correctly and register features.
+
+:::warning
+**This module should almost never be configured directly in project code.** In other words, there should almost never be a `modules/@apostrophecms/module/index.js` file in a project codebase or `@apostrophecms/module: {}` line in an `app.js` file. The only reason to do so would be to add a feature that every other module will need.
+
+**Instead, configure the options below on other modules as needed** (e.g., configure the `alias` option on a module to create an easy reference to *that* module). Similarly, run the methods below as needed from the most appropriate module (e.g., use `self.email` from a contact page-type module).
+:::
+
+## Options
+
+|  Property | Type | Description |
+|---|---|---|
+|`alias` | String | A name to use for quick access to the module on the `apos` object (e.g., `'book'` to access a module on `self.apos.book`). Otherwise the longer syntax is necessary (e.g., `self.apos.modules.book`). |
+|`templateData` | Object | An object of properties to include on the `data` object in templates belonging to that module. |
+
+## Featured methods
+
+The following methods belong to this module and may be useful in project-level code. As noted above, *they should be used via another, more appropriate module,* rather than directly called from this module. See the [source code](https://github.com/apostrophecms/apostrophe/blob/main/modules/%40apostrophecms/module/index.js) for all methods that belong to this module.
+<!-- Some are used within the module and would just create noise here. -->
+
+### `async render(req, template, data)`
+
+Returns rendered HTML for a template with the data provided. You must pass `req`, a request object, as the first argument. The `template` argument should be the name of a template file in the module's `views` directory.
+
+All properties of `data` can be used in Nunjucks templates as properties of the `data` object. This argument may be omitted to include no additional data.
+
+### `async renderString(req, string, data)`
+
+Returns rendered HTML for a Nunjucks-style string with the data provided. You must pass `req`, a request object, as the first argument. The `string` argument will be used as the template itself for rendering (not the template filename).
+
+All properties of `data` can be used in Nunjucks templates as properties of the `data` object. This argument may be omitted to include no additional data.
+
+### `async send(req, template, data)`
+
+The `send()` method renders a template with data as with the [`render()`](#async-render-req-template-data), then sends the rendered HTML as a response to the request (`req`).
+
+### `async sendPage(req, template, data)`
+
+Similarly to the [`send()`](#async-send-req-template-data) method renders a template and sends the rendered HTML as a response to the request (`req`). Where `send()` is used to render general template files, `sendPage()` is specifically used to render and send *full pages* for Apostrophe projects.
+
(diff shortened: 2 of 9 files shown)


docs/reference/modules/_template.md
```diff
@@ -1,9 +1,15 @@
+---
+extends: '@apostrophecms/module'
```
abea, line 2: I'm using frontmatter since it seems likely that we'll want to have more programmatic access to this at some point.

docs/reference/modules/_template.md
```diff
@@ -27,7 +33,7 @@ The following methods belong to this module and may be useful in project-level c
 
 Because this module has an alias, you can call these from another module from the alias path. For example, `self.apos.[the alias].inferIdLocaleAndMode()`.
 
-### `inferIdLocaleAndMode(req, _id)`
+### `async inferIdLocaleAndMode(req, _id)`
```
abea, line 36: Adding this as a reminder to identify async methods this way. I realize this particular method isn't async, but this is simply the template.

docs/reference/server-events.md
```diff
@@ -673,6 +673,29 @@ There is no data included with the event for handlers.
 
 Events emitted by the `@apostrophecms/page` module.
 
+### `beforeSend`
```
abea, line 676: I'm not sure how I missed this one before.

docs/reference/modules/module.md
```diff
@@ -0,0 +1,103 @@
...
+## Options
+
+|  Property | Type | Description |
+|---|---|---|
+|`alias` | String | A name to use for quick access to the module on the `apos` object (e.g., `'book'` to access a module on `self.apos.book`). Otherwise the longer syntax is necessary (e.g., `self.apos.modules.book`). |
+|`templateData` | Object | An object of properties to include on the `data` object in templates belonging to that module. |
+
+## Featured methods
+
+The following methods belong to this module and may be useful in project-level code. As noted above, *they should be used via another, more appropriate module,* rather than directly called from this module. See the [source code](https://github.com/apostrophecms/apostrophe/blob/main/modules/%40apostrophecms/module/index.js) for all methods that belong to this module.
```
boutell, line 20: They cannot be directly called on this module, since it is never instantiated, it is just an abstract base class.

docs/reference/modules/module.md
```diff
@@ -0,0 +1,103 @@
...
+<!-- Some are used within the module and would just create noise here. -->
+
+### `async render(req, template, data)`
+
+Returns rendered HTML for a template with the data provided. You must pass `req`, a request object, as the first argument. The `template` argument should be the name of a template file in the module's `views` directory.
```
boutell, line 25: If there is no file extension, Apostrophe will automatically look for a `.njk` or `.html` file.

docs/reference/modules/module.md
```diff
@@ -0,0 +1,103 @@
...
+
+All properties of `data` can be used in Nunjucks templates as properties of the `data` object. This argument may be omitted to include no additional data.
+
+### `async send(req, template, data)`
+
+The `send()` method renders a template with data as with the [`render()`](#async-render-req-template-data), then sends the rendered HTML as a response to the request (`req`).
+
+### `async sendPage(req, template, data)`
+
+Similarly to the [`send()`](#async-send-req-template-data) method renders a template and sends the rendered HTML as a response to the request (`req`). Where `send()` is used to render general template files, `sendPage()` is specifically used to render and send *full pages* for Apostrophe projects.
```
boutell, line 41: Grammar.

boutell: There are some pretty critical details in the inline documentation of this method. In particular the developer is meant to use:

`{% extends data.outerLayout %}`

In their template, otherwise calling this method doesn't accomplish much over `send`.

The inline documentation in the source is wrong on this point, it leaves out the `data.` and says `extend` rather than `extends`.

docs/reference/modules/module.md
```diff
@@ -0,0 +1,103 @@
...
+
+### `enableBrowserData(scene)`
+
+Call this method from a module's [initialization function](/reference/module-api/module-overview.md##initialization-function) if the module implements the [`getBrowserData()` method](#getbrowserdata-req). **This is only necessary for modules that do not already do this** (all doc type and widget type modules already do this). The data returned by `getBrowserData(req)` will then be available on `apos.modules['the-module-name']` in the browser.
+
+The `scene` argument is only needed if making the data available while logged-out. By default browser data is pushed only for the `apos` scene (the logged-in context), so the anonymous site visitor experience will include the extra data (except on the /login page and other pages that opt into the `apos` scene). If `scene` is set to `public` then the data is available to all visitors.
+
+### `getBrowserData(req)`
+
+The `getBrowserData()` method can be extended to adjust or add data that the module makes available in the browser. It should return an object to make the data available in browsers (as controlled by [the `enableBrowserData()` method](#enablebrowserdata-scene)).
```
boutell, line 55: Feel like we need a good mnemonic indicator for "this is a method you implement, not a method you call."

docs/reference/modules/module.md
```diff
@@ -0,0 +1,103 @@
...
+
+Many Apostrophe core modules already populate browser data, including piece type, page type, and widget type modules. For this reason, **it is ususally correct to *extend* this method to add additional browser data** using the module's [`extendMethods` customization function](/reference/module-api/module-overview.md#extendmethods-self).
+
+Avoid returning large data structures, as this will impact page load time and performance.
+
+### `getOption(req, dotPathOrArray, def)`
+
+This is a convenience method to fetch properties of `self.options` in a module. It will look for, then return the option identified by the second argument.
+
+`req` is required to provide extensibility. Modules can use it to change the response based on the current page and other factors tied to the request. The second argument, `dotPathOrArray`, may be a dot path (`'flavors.grape.sweetness'`) or an array `[ 'flavors', 'grape', 'sweetness' ]`.
```
boutell, line 67: Modules that override this method can use it...

boutell: Actually I think we're going to deprecate use of req in `getOption`. It underpins apostrophe-override-options in A2 but we've successfully advocated so far for replacing that with just a way for the global doc to inject options into modules, which would happen by recreating the apos object like in multisite, so no need for this messiness.

boutell: Since the key is always a string we should be able to introduce a "just pass the key and don't pass req at all" variation too once we commit on that.

docs/reference/server-events.md
```diff
@@ -673,6 +673,29 @@ There is no data included with the event for handlers.
...
+
+Triggered just before a page template is rendered and sent as a response to a request. This can be a good place to adjust the data available to the template  by amending `req.data`.
```
boutell, line 678: See also async components, which can be a cleaner way to package code that fetches additional data during page rendering.

boutell: (I'm not suggesting we should deprecate or not document beforeSend. It still has its place I think. Just a little hard to nail down exactly when that is when async components are also available. We'll let best practices evolve)

docs/reference/modules/_template.md
```diff
@@ -1,9 +1,15 @@
...
+extends: '@apostrophecms/module'
+---
+
 # `@apostrophecms/module-name`
 
 **Alias:** `apos.i18n`
 
 *General description paragraph.* Cras mattis consectetur purus sit amet fermentum. Lorem ipsum dolor sit amet, consectetur adipiscing elit.
 
+**Extends:** `{{ $frontmatter.extends }}`
```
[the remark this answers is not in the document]
abea: I don't really think this will be a problem for that. Most devs won't be in these docs often and code examples will guide them the correct way. We use "extends" elsewhere:
> Inheritance is the glue of the module system. Every module extends another module, inheriting functionality and structure. This means that your blog post module, which extends the ["piece type"](/reference/glossary.md#piece) module, comes with a huge set of features you never have to write.

If you think it's a problem though, I can use "inherits from," like A2, though it feels clunkier to me.

docs/reference/modules/module.md
abea: Yeah, I realized a bit into this one how that's a real distinction here. This would be a good case for a code recipe article.

docs/reference/modules/module.md
abea: > Modules that override this method can use it...

I'm not sure what that means.

abea: > Actually I think we're going to deprecate use of req in getOption.

We can remove this for now, comment it out, or leave it for and update when that's deprecated. What do you prefer?

boutell (approved):

PR #224, opened 2022-07-25
Pro 2939 piece page type doc

BoDonkey:
## Summary

This PR adds a new reference page documenting the piece-page-type. Closes PRO-2939.

*For example:*
1. Build the documentation
2. Navigate to reference->modules->piece-page-type.html

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [X] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/docs/reference/modules/piece-page-type.md
+++ b/docs/reference/modules/piece-page-type.md
@@ -6,278 +6,235 @@ extends: '@apostrophecms/doc-type'
 
 <AposRefExtends :module="$frontmatter.extends" />
 
-This module operates in cooperation with the 'piece-type' module to show both individual pieces, as well as all instances of a particular piece. In addition to all of the features from the ['@apostrophecms/page'](/reference/modules/page.md) module, it also exposes two specialized views and additional data to the templates.
-
-TODO: Expand a bit?
+This module cooperates with the 'piece-type' module to expose two specialized views. The [index page](https://v3.docs.apostrophecms.org/reference/glossary.html#index-page) displays all pieces of a particular `piece-type` in a paginated, filterable manner. The [show page](https://v3.docs.apostrophecms.org/reference/glossary.html#show-page) is for presenting individual pieces. These features are added to those exposed by the  ['@apostrophecms/page'](/reference/modules/page.md) module.
 
 ## Options
 
 |  Property | Type | Description |
 |---|---|---|
-| [`label`](#label) | String | The human-readable label for the doc type. |
-| [`next`](#next) | Boolean | If set to `true`, `data.next` is the next piece based on the sort. |
+| [`next`](#next) | Boolean \|\| Object | If set to `true`, `data.next` is the next piece based on the sort. |
 | [`perPage`](#perpage) | Integer | The number of pieces to include in a set of `GET` request results. |
-| [`piecesFilters`](#autopublish) | Boolean | Set to `true` to publish all saved edits immediately. |
+| [`piecesFilters`](#piecesFilter) | Array | Takes an array of objects where each contains a `name` key and a value of a field in the piece to filter on. |
+| [`pieceModuleName`](#pieceModuleName) | String | Optionally sets the `piece-type` to a specific name other than the default set by the module name. |
+| [`previous`](#previous) | Boolean \|\| Object | If set to `true`, `data.previous` is the previous piece based on the sort. |
 
-### `autopublish`
+### `next`
+If this option is set to true, it exposes the next piece in the current [sort order](https://v3.docs.apostrophecms.org/reference/module-api/module-options.html#sort) as `req.data.next` (`data.next` in the template) when serving a [show page](https://v3.docs.apostrophecms.org/reference/glossary.html#general-terms). This can be used to provide a link to the next item in a series (e.g., the next oldest blog post).
 
-Set `autopublish` to `true` to automatically publish any changes saved to docs of this type. There is then effectively no draft mode for this doc type, but there will be draft document versions in the database.
+This option can also be set to an object containing an object with a key of `project` and values which will be used as a query builder for retrieving the next piece document.
 
-The core image and file modules use this option, for example. It eliminates the need for users to think about the distinction between draft and published content while preserving the possibility of translation for different locales.
-
-#### Example
+<AposCodeBlock>
 
 ```javascript
-// modules/article-category/index.js
 module.exports = {
-  extend: '@apostrophecms/piece-type',
+  extend: ‘@apostrophecms/piece-page-type’,
   options: {
-    autopublish: true
+    next: true
   },
-  // ...
+  // …
 }
-```
 
-### `cache`
+// OR
 
-`cache` can be set to an object with an `api` subproperty, and a `maxAge` subproperty within that, determining the cache lifetime in seconds. If enabled, Apostrophe will send a `Cache-Control` header with the specified maximum age. The actual caching is provided by the browser, or by an intermediate CDN or reverse proxy.
-
-Note that Apostrophe already provides "cache on demand" by default, to improve performance when simultaneous `GET` requests arrive for the same piece. Unlike "cache on demand," setting the `cache` option introduces the possibility that some visitors will see older content, up to the specified lifetime.
-
-If a user is logged in, or `req.session` has content, Apostrophe always disables caching. However such a user could encounter a previously cached document from before logging in. Apostrophe contains logic to mitigate this in the editing experience.
-
-#### Example
-
-```javascript
-  cache: {
-    api: {
-      // Specified in seconds
-      maxAge: 3000
+module.exports = {
+  extend: '@apostrophecms/piece-page-type',
+  options: {
+    // The next article piece would be returned with only the `title, `_url`
+    // and `lastPublishedAt` properties.
+    next: {
+      project: {
+        title: 1,
+        _url: 1,
+        lastPublishedAt: 1
+      }
     }
-  }
-```
-
-### `label`
-
-`label` should be set to a text string to be used in user interface elements related to this doc type. This includes buttons to open piece manager modals.
-
-If not set, Apostrophe will convert the module name to a readable label by splitting the `name` property on dashes and underscores, then capitalizing the first letter of each word.
-
-#### Example
-
-```javascript
-// modules/feature/index.js
-module.exports = {
-  extend: '@apostrophecms/piece-type',
-  options: {
-    label: 'Featured Article'
   },
-  // ...
+  // …
 }
 ```
-
-### `localized`
-
-Defaults to `true`. If set to `false`, this doc type will _not_ be included in the locale system. This means there will be only one version of each doc, regardless of whether multiple locales (e.g., for languages or regions) are active. There is no distinction between draft and published, including in the database.
-
-The "users" piece type disables localization in this way. It can also be useful for piece types that are synchronized from another system that has no notion of locales and no distinction between "draft" and "published" content.
-
-#### Example
-
-```javascript
-// modules/administrative-category/index.js
-module.exports = {
-  extend: '@apostrophecms/piece-type',
-  options: {
-    localized: false
-  },
-  // ...
-}
-```
+<template v-slot:caption>
+  modules/article-page/index.js
+</template>
+</AposCodeBlock>
 
 ### `perPage`
+The `perPage` option receives an integer as value and sets the number of pieces that will be displayed per page for the `index.html` page. It is set to 10 items per page by default.
 
-In piece types, the `perPage` option, expressed as an integer, sets the number of pieces that will be returned in each "page" [during `GET` requests](/reference/api/pieces.md#get-api-v1-piece-name) that don't specify an `_id`. It also controls how many are displayed in the manager modal user interface. This value defaults to 10.
-
-#### Example
+<AposCodeBlock>
 
 ```javascript
-// modules/article/index.js
 module.exports = {
   extend: '@apostrophecms/piece-type',
   options: {
-    perPage: 20 // REST `GET` requests will return 20 pieces per page.
+    perPage: 20 // REST `GET` requests will return 20 pieces per page
   },
-  // ...
+// …
 }
 ```
+  <template v-slot:caption>
+    modules/article-page/index.js
+  </template>
+</AposCodeBlock>
 
-### `pluralLabel`
+### `piecesFilter`
+The `piecesFilter` takes an array of objects to assist in filtering on the index page. Each object must have a `name` property associated with a valid [query builder](https://v3.docs.apostrophecms.org/reference/module-api/module-overview.html#queries-self-query). 
 
-Similar to `label` for all doc types, the `pluralLabel` option sets the string the user interface will use to describe a piece type in plural contexts.
+These include:
+* Custom query builders configured in an app that include a `launder` method
+* Field names whose field types automatically get builders:
+    - boolean
+    - checkboxes
+    - date
+    - float
+    - integer
+    - relationship
+    - select
+    - slug
+    - string
+    - url
 
-If no `pluralLabel` value is provided, Apostrophe will append the `label` (whether set manually or generated [as described](#label)), with "s", as is typical for English words. **Even in English this is often not correct, so `pluralLabel` should usually be defined explicitly.**
+When the index page is served, configured filters will be represented on a `req.data.piecesFilter` object (`data.piecesFilter` in the template). If you include `counts: true` in a filter object the number of pieces matching that filter are included in the `req.data.piecesFilter` properties.
 
-#### Example
+<AposCodeBlock>
 
 ```javascript
-// modules/goose/index.js
 module.exports = {
   extend: '@apostrophecms/piece-type',
   options: {
-    label: 'Goose',
-    pluralLabel: 'Geese'
-  },
-  // ...
-}
-```
-
-### `publicApiProjection`
-
-By default the built-in Apostrophe REST APIs are not accessible without proper [authentication](/reference/api/authentication.md). You can set an exception to this for `GET` requests to return specific document properties with the `publicApiProjection` option.
-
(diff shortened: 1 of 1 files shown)


docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+This module cooperates with the 'piece-type' module to expose two specialized views. The [index page](https://v3.docs.apostrophecms.org/reference/glossary.html#index-page) displays all pieces of a particular `piece-type` in a paginated, filterable manner. The [show page](https://v3.docs.apostrophecms.org/reference/glossary.html#show-page) is for presenting individual pieces. These features are added to those exposed by the  ['@apostrophecms/page'](/reference/modules/page.md) module.
+
+## Options
+
+|  Property | Type | Description |
+|---|---|---|
+| [`next`](#next) | Boolean \|\| Object | If set to `true`, `data.next` is the next piece based on the sort. |
+| [`perPage`](#perpage) | Integer | The number of pieces to include in a set of `GET` request results. |
+| [`piecesFilters`](#piecesFilter) | Array | Takes an array of objects where each contains a `name` key and a value of a field in the piece to filter on. |
+| [`pieceModuleName`](#pieceModuleName) | String | Optionally sets the `piece-type` to a specific name other than the default set by the module name. |
```
boutell, line 18: inferred from

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+| [`previous`](#previous) | Boolean \|\| Object | If set to `true`, `data.previous` is the previous piece based on the sort. |
+
+### `next`
+If this option is set to true, it exposes the next piece in the current [sort order](https://v3.docs.apostrophecms.org/reference/module-api/module-options.html#sort) as `req.data.next` (`data.next` in the template) when serving a [show page](https://v3.docs.apostrophecms.org/reference/glossary.html#general-terms). This can be used to provide a link to the next item in a series (e.g., the next oldest blog post).
+
+This option can also be set to an object containing an object with a key of `project` and values which will be used as a query builder for retrieving the next piece document.
```
boutell, line 24: It can be set to an object whose keys are query builders, such as `project`, and whose values are the parameters to be passed to each query builder.

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+
+<AposCodeBlock>
+
+```javascript
+module.exports = {
+  extend: ‘@apostrophecms/piece-page-type’,
```
boutell, line 30: Wrong quotes.

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+  // …
+}
+```
+<template v-slot:caption>
+  modules/article-page/index.js
+</template>
+</AposCodeBlock>
+
+### `perPage`
+The `perPage` option receives an integer as value and sets the number of pieces that will be displayed per page for the `index.html` page. It is set to 10 items per page by default.
```
boutell, line 61: should be set to an integer

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+
+<AposCodeBlock>
+
+```javascript
+module.exports = {
+  extend: '@apostrophecms/piece-type',
+  options: {
+    perPage: 20 // REST `GET` requests will return 20 pieces per page
```
boutell, line 69: This isn't about REST.

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+    - date
+    - float
+    - integer
+    - relationship
+    - select
+    - slug
+    - string
+    - url
+
+When the index page is served, configured filters will be represented on a `req.data.piecesFilter` object (`data.piecesFilter` in the template). If you include `counts: true` in a filter object the number of pieces matching that filter are included in the `req.data.piecesFilter` properties.
```
boutell, line 96: Spell out what the values inside `piecesFilter` will look like.

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+```
+<template v-slot:caption>
+modules/fiction-page/index.js
+</template>
+</AposCodeBlock>
+
+### `previous`
+If this option is set to true, it exposes the previous piece in the current [sort order](https://v3.docs.apostrophecms.org/reference/module-api/module-options.html#localized) as `req.data.previous` (`data.previous` in the template) when serving a [show page](https://v3.docs.apostrophecms.org/reference/glossary.html#general-terms). This can be used to provide a link to the previous item in a series (e.g., the next newest blog post).
+
+This option can also be set to an object containing an object with a key of `project` and values which will be used as a query builder for retrieving the previous piece document.
```
boutell, line 177: See earlier comment. `project` is an example, not the only possible use (a useful one though, for performance).

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+## Related Documentation
+
+* [Piece index and show pages](https://v3.docs.apostrophecms.org/guide/piece-pages.html)
+* [Piece page type options](https://v3.docs.apostrophecms.org/reference/module-api/module-options.html#options-for-piece-page-types)
+
+## Featured methods
+The following methods belong to this module and may be useful in project-level code. See the [source code](https://github.com/apostrophecms/apostrophe/blob/main/modules/%40apostrophecms/piece-page-type/index.js) for all the modules that belong to this module.
+
+### `indexQuery(req)`
+This method should be overridden for a piece-type to call additional [query builders](https://v3.docs.apostrophecms.org/reference/query-builders.html#query-builders) by default.
```
boutell, line 222: replace "by default" with "when generating the index page."

The base class version of the method returns a query object. Overriding looks like:

```
extendMethods(self) {
  return {
    indexQuery(_super, req) {
      // Works if there is a select field named "category"
      return _super(req).category('fiction');
    }
  };
}
```

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+
+### `beforeIndex(req)`
```
boutell, line 224: `async beforeIndex(req)`

Overrides can use `await`. Also true for `beforeShow`

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+This method is called before `indexPage`. It is a convenient method to extend for manipulating the `req` being supplied to that page.
+
+### `beforeShow(req)`
+This method is called before `showPage`. It is a convenient method to extend for manipulating the `req` being supplied to that page.
+
+### `dispatchAll()`
+This method can be extended to override the default behavior of invoking `showPage` if the URL has an additional path after the base, e.g. `/blog/good-article`. As example, you could override to use `/:year/:month/:day/:slug` to invoke `self.showPage`. This should be used in conjunction with the [`buildUrl()`](#buildurl-req-page-piece)) method of this module. See [@apostrophecms/page-type](https://github.com/apostrophecms/apostrophe/blob/main/modules/%40apostrophecms/page-type/index.js) for more about what you can do with dispatch routes.
```
boutell, line 231: Please make a note to update this when reference material for `page-type` has been completed or a guide page specifically for `dispatch` has been written. You don't have to wedge it into this PR though, I want to avoid making individual tickets insurmountable.

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+
+### `buildUrl(req, page, piece)`
+This method should be extended to build custom URL for use in the [`dispatchAll`](#dispatchall) method.
```
boutell, line 234: I would suggest "this method can be overridden to change the URLs that are generated as the `_url` property for individual pieces. Note that the `dispatchAll` method often must also be overridden to ensure those URLs actually reach those pieces."

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+
+### `filterByIndexPage(query, page)`
+This method invokes query builders on the supplied query argument to ensure it only fetches results appropriate to the given page. This is typically done when there is more than one pieces-page per page type. This should be used in conjunction with the [`chooseParentPage`](#chooseparentpage-pages-piece) method of this module.
```
boutell, line 237: Please make clear that this method does nothing by default, and should be overridden to have the effect described in use cases where it is needed. Also note this clearly for other methods mentioned here that are empty in core.

Overriding and extending are notably different in Apostrophe because the easiest way to override (completely replace) a method is to just write it in your own module's `methods` section, while the easiest way to extend a method (invoke the original, but return a modified result, etc) is to use the module's `extendMethods` section. So we should use the word "override" specifically when a method is initially empty by design and you don't need to worry about using `extendMethods`.

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
...
+
+### `chooseParentPage(pages, piece)`
+This `pages` parameter of this method takes an array of all of the index pages for a particular piece-type, and the `piece` parameter is an individual piece-type name. By default, it will return the first item in the `pages` array, but the developer should map pieces to pages in a way that makes sense for their design. This method will give a warning if the [`filterByIndexPage`](#filterbyindexpage-query-page) method has not been overridden.
```
boutell, line 240: This -> The

"The default version of this method will give a warning if it sees more than one page in the array, as it is up to you to override this method to provide a sensible way of deciding which page is the best 'parent' for each piece."

It's not about whether `filterByIndexPage` has been overridden, the warning just mentions that it's a good idea to override that too in the same situation, so each index page displays the same pieces that would logically regard it as a "parent."

docs/reference/modules/piece-page-type.md
```diff
@@ -0,0 +1,240 @@
+---
+extends: '@apostrophecms/doc-type'
+---
+
+# `@apostrophecms/piece-page-type`
+
+<AposRefExtends :module="$frontmatter.extends" />
+
+This module cooperates with the 'piece-type' module to expose two specialized views. The [index page](https://v3.docs.apostrophecms.org/reference/glossary.html#index-page) displays all pieces of a particular `piece-type` in a paginated, filterable manner. The [show page](https://v3.docs.apostrophecms.org/reference/glossary.html#show-page) is for presenting individual pieces. These features are added to those exposed by the  ['@apostrophecms/page'](/reference/modules/page.md) module.
```
boutell, line 9: Once an editor adds a page of this type to the site via user interface, it becomes possible to view a listing of pieces by visiting that page's URL, and to view individual pieces of the relevant type by adding the slug of any piece to the page's URL, like this: `/slug/of/page/slug-of-piece`

This default behavior can be customized as described below.

It is possible to add more than one such page to the site for the same type, and to add custom logic to decide which pieces should be associated with each such page.

boutell (changes requested):

docs/reference/modules/piece-page-type.md
BoDonkey: Changed

docs/reference/modules/piece-page-type.md
BoDonkey: Changed.

docs/reference/modules/piece-page-type.md
BoDonkey: I should have noticed that. As an aside - I'm not completely positive how I made what I assume is a forward tick.

docs/reference/modules/piece-page-type.md
BoDonkey: Changed and slightly edited for word choice.

docs/reference/modules/piece-page-type.md
BoDonkey: Comment deleted. I got this example from the `piece-type` reference page and copied without enough thought. This does spark a technical question on what really is happening on the back end in the app.

docs/reference/modules/piece-page-type.md
BoDonkey: Reworked this paragraph and added an example of what `data.piecesFilters` will look like.

docs/reference/modules/piece-page-type.md
BoDonkey: Also realized that the `extend` value was incorrect.

docs/reference/modules/piece-page-type.md
BoDonkey: See earlier response.

docs/reference/modules/piece-page-type.md
BoDonkey: Should I include this example?

docs/reference/modules/piece-page-type.md
BoDonkey: Changed for both.

docs/reference/modules/piece-page-type.md
BoDonkey: Done.

docs/reference/modules/piece-page-type.md
BoDonkey: Changed for `filterByIndexPage`, `beforeIndex`, and `beforeShow`.

docs/reference/modules/piece-page-type.md
BoDonkey: Changed. I have to think about this one a little more for my understanding.

docs/reference/modules/piece-page-type.md
BoDonkey: Added.

docs/reference/modules/piece-page-type.md
```diff
@@ -8,26 +8,32 @@ extends: '@apostrophecms/doc-type'
 
 This module cooperates with the 'piece-type' module to expose two specialized views. The [index page](https://v3.docs.apostrophecms.org/reference/glossary.html#index-page) displays all pieces of a particular `piece-type` in a paginated, filterable manner. The [show page](https://v3.docs.apostrophecms.org/reference/glossary.html#show-page) is for presenting individual pieces. These features are added to those exposed by the  ['@apostrophecms/page'](/reference/modules/page.md) module.
 
+Once an editor adds a page of this type to the site via user interface, it becomes possible to view a listing of pieces by visiting that page's URL, and to view individual pieces of the relevant type by adding the slug of any piece to the page's URL, like this: `/slug/of/page/slug-of-piece`
```
boutell, line 11: via *the* user interface

docs/reference/modules/piece-page-type.md
```diff
@@ -58,15 +64,15 @@ module.exports = {
 </AposCodeBlock>
 
 ### `perPage`
-The `perPage` option receives an integer as value and sets the number of pieces that will be displayed per page for the `index.html` page. It is set to 10 items per page by default.
+The `perPage` option should be set to an integer as a value and specifies the number of pieces displayed per page for the `index.html` page. It is set to 10 items per page by default.
```
boutell, line 67: "as a value" is not adding anything here I think.

I'dd add "before pagination is offered."

docs/reference/modules/piece-page-type.md
```diff
@@ -142,13 +148,33 @@ module.exports = {
...
     modules/book-page/index.js
   </template>
 </AposCodeBlock>
 
+<AposCodeBlock>
+
+```
+{
+  _author: [
+    { value: 'Gibson', label: 'Gibson' },
```
boutell, line 162: `_author` would be a relationship field, so the values are going to be IDs (as in `._id` properties of docs), not user-friendly strings.

boutell: The labels will be friendly of course.

docs/reference/modules/piece-page-type.md
```diff
@@ -93,7 +99,7 @@ These include:
     - string
     - url
 
-When the index page is served, configured filters will be represented on a `req.data.piecesFilter` object (`data.piecesFilter` in the template). If you include `counts: true` in a filter object the number of pieces matching that filter are included in the `req.data.piecesFilter` properties.
+When the index page is served, filter data will be returned in the `req.data.piecesFilters` object (`data.piecesFilters` in the template). This object consists of an array for each configured filter. That array contains objects with `value` and `label` properties for every `piece-type` that matches the filter. If `counts: true` is included for the filter query, each object in the array will also have a `count` property with the number of matching pieces.
```
boutell, line 102: Passing any of these filters back to the index page as query string parameters will filter the results accordingly. <-- This is the actual win of having the filters, so we should spell it out. An example using `| build` would be nice, but maybe more appropriate for a new "guide" advanced page about these features, which currently are documented only here and in the "module options" reference page .

boutell (changes requested):

docs/reference/modules/piece-page-type.md
BoDonkey: Changed.

docs/reference/modules/piece-page-type.md
BoDonkey: Changed.

docs/reference/modules/piece-page-type.md
BoDonkey: Changed for output from a relationship field filter. I was going to change it to a simple `string` type, but I thought this would be a good example.

docs/reference/modules/piece-page-type.md
BoDonkey: Changed. I should have thought to include this. It was one of my 'AhhHaa!' moments working on this documentation. I didn't know this fact and suddenly discovered this fact and a LOT of things gelled in my head. Not sure if you want the `| build` example here or not - up to you.

boutell (approved):

boutell:
Approved. `| build` examples would make a nice follow-on but might be more appropriate in a guide section article.

PR #235, opened 2022-08-12
Pro 3079 api options

BoDonkey:
## Summary

The reference docs for the API incorrectly had the query parameters as kebab-case, rather than camelCase. This PR corrects this. It also changes the example usage to reflect adding parameter queries rather than altering the passed id. Closes PRO-3097

*For example:*

1. Build the documentation
2. Navigate to either `reference/api/pieces.html` or `/reference/api/pages.html`
3. Confirm that all of the query params are camelCase.
4. 
*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/docs/reference/api/pages.md
+++ b/docs/reference/api/pages.md
@@ -217,6 +217,7 @@ Individual page objects will include `_children` and `_ancestor` arrays, as well
 <!-- Read more about [mode and locale parameters on single-document requests](/guide/rest-apis#locale-and-mode-in-single-document-requests). -->
 ::: note
 Query parameters will override the locale and mode present in the `_id`. So, if the `aposLocale=es` parameter is supplied, a `GET` request to the `_id` `###:en:published` will return the Spanish, not English, locale.
+You can also elect to use the `aposDocId` instead of the `_id` and use the query parameters to pass in the locale and mode parameters found in the `_id`.
 :::
 ### Request example
 
@@ -309,7 +310,7 @@ const data = {
   _position: 'lastChild'
 };
 // Request inside an async function.
-const response = await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr?apikey=myapikey&aposMode=published&aposLocale=fr', {
+const response = await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr:fr:published?apikey=myapikey', {
   method: 'PUT',
   headers: {
     'Content-Type': 'application/json'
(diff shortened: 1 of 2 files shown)


docs/reference/api/pages.md
```diff
@@ -220,7 +220,7 @@ Individual page objects will include `_children` and `_ancestor` arrays, as well
 
 ```javascript
 // Request inside an async function.
-const response = await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr:en:published?apikey=myapikey', {
```
boutell, line 223: Wait, you lost me here. If you have a complete `_id` you may send it and that information is sufficient to ensure you get the right locale and mode. The query parameters are for situations where there's no `_id` or you want to override some part of what it implies.

docs/reference/api/pages.md
```diff
@@ -220,7 +220,7 @@ Individual page objects will include `_children` and `_ancestor` arrays, as well
...
+const response = await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr?apikey=myapikey&aposMode=published&aposLocale=en&renderAreas=true', {
```
boutell, line 223: This is a great example too, but a separate one. The more common case is "I've got an _id, probably because I did a get all operation, and now I want a doc back with all the trimmings, and I don't want to think about modes and locales when that stuff is all in the _id already."

boutell: It's true that you can send `aposDocId` instead of `_id` to leave off the current locale and mode of the document you already have, and specify the rest via query params (which always win if the info is present in the id), but that's not obvious nor is it being explained here.

boutell (changes requested):

docs/reference/api/pages.md
BoDonkey: Corrected back to the original.

docs/reference/api/pages.md
BoDonkey: I added a sentence about the parameters overriding, but I can remove it.
I just adapted the existing examples to use query params instead of the `_id` to illustrate the usage.

docs/reference/api/pages.md
```diff
@@ -307,7 +309,7 @@ const data = {
   _position: 'lastChild'
 };
 // Request inside an async function.
-const response = await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr:en:published?apikey=myapikey', {
+const response = await fetch('http://example.net/api/v1/@apostrophecms/page/ckitdo5oq004pu69kr6oxo6fr?apikey=myapikey&aposMode=published&aposLocale=fr', {
```
boutell, line 312: If you are going to show examples using `aposDocId` instead of the complete `_id` you need to show that so developers know where to get it from.

Also we don't want to lose all examples showing it's perfectly fine to use the plain `_id` if you're happy with the locale and mode in it.

boutell (changes requested):

docs/reference/api/pages.md
BoDonkey: I changed a few back and expanded upon the note within the topmost GET that takes `_id`. Wasn't sure if it would be redundant to add that same message in multiple areas.

boutell (approved):

PR #292, opened 2023-02-01
docs for images in rich text

boutell:
Let's get this ready before we ship a3 core


BoDonkey (commented):
In general, this looks good, but I have a potential clarification question. You state that it adds a `figcaption` element. Is that element populated from the image alt text, or...? Should we add a statement about this?

PR #331, opened 2023-05-25
Pro 4185 webhook recipe

BoDonkey:
## Summary

This PR adds a recipe for creating webhooks. It came out of the Bejamas comparison between the feature of Apostrophe and other CMS. We needed an article to show equality. Closes PRO-4185.

*For example:*

1. Build the documentation
2. Navigate to `/cookbook/creating-webhooks.html`

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [X] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,105 @@
+# Creating webhooks
+
+Webhooks allow applications to communicate with each other in real time by sending HTTP requests when certain events or triggers occur. This can allow your site to automatically send notifications when publishing a document or receive updates from payment portals, for example. This short recipe will demonstrate how to set up webhooks in your Apostrophe project.
+
+## Sending information to a 3rd party webhook
```
boutell, line 5: "Outgoing webhooks" is a common term of art for this. Should be here for SEO reasons and because it's familiar.

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,105 @@
...
+  return {
+    'afterPublish': {
+      async sendWebhook(req, data) {
+        // data sent by the afterPublish event includes the document data 
+        // and whether it is being published for the first time
+        // https://v3.docs.apostrophecms.org/reference/server-events.html#afterpublish
+        if (!data.firstTime) {
+          return;
+        }
+        // The url added here should be changed to the desired endpoint
```
boutell, line 22: This is OK, but I think posting to slack would be almost as simple and a true example.

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,105 @@
...
+
+The next section of code is involved in setting up the HTTP `POST` request and will depend heavily on what the desired endpoint requires, and what content you want sent to the endpoint. In this example, we are adding the piece `title` and `_id`.
+
+Finally, we are using the `post` method of the [`@apostrophecms/http` module](https://v3.docs.apostrophecms.org/reference/modules/http.html#async-post-url-options) to send our data to the endpoint. Depending on the endpoint, the returned response might be as simple as a `200` success or a `400` failed response, or might contain additional data. That response can be handled to retry a failed response or in some way log the returned data.
+
+## Setting up a webhook endpoint
```
boutell, line 55: "Incoming webhooks" is the frequently seen term of art in this direction.

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,105 @@
...
+
+Your project may require a webhook endpoint for interfacing with a payment or subscription service. This can be accomplished by using the Apostrophe [`apiRoutes(self)` module configuration](https://v3.docs.apostrophecms.org/reference/module-api/module-overview.html#apiroutes-self) function. This allows you to easily set up an endpoint for any HTTP request method.
```
boutell, line 57: This paragraph could do more to express this is about accepting incoming connections.

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,105 @@
...
+
+```javascript
+options: {
+  csrfExceptions: [ '/webhooks' ]
+},
+apiRoutes(self) {
+  return {
+    post: {
+      // change function name to match the desired endpoint URL
+      '/webhooks': async function(req) {
```
boutell, line 69: I would suggest not rooting this with a `/` in the example, but instead explaining it'll automatically be `/api/v1/your-module-name/your-method-name` for tidy namespacing, unless you do start it with a `/`. Usually services to which you provide a webhook like this allow you to paste the URL you want.

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,105 @@
...
+        // a specific type of authorization
+        const authHeader = req.headers.authorization;
+        if (!authHeader) {
+          throw self.apos.error('forbidden');
+        }
+
+        // Use data passed by webhook to update database
+        // or provide some other type of functionality
+        const webhookData = req.body;
+        addNewSubscriber(webhookData);
```
boutell, line 83: `await` this. The big win with apiRoutes is that they automatically handle exceptions for you and allow you to return values you got from async functions, but not if you don't await them. Then you just have a race condition.

docs/cookbook/creating-webhooks.md
boutell: If you just put the function name in `csrfExceptions` (without the `/`) it will do the same math to figure out the URL.

I can see how this is kind of a lot when I spell it all out. Maybe you're right to just use a rooted example with `/`.

docs/cookbook/creating-webhooks.md
boutell:
```
await self.apos.http.post('https://slack-webhook-url-here', {
  body: {
    text: `do something cool with the data`,
  }
});
```

Adapted from what's in our assembly cloud tools.

boutell (changes requested):

docs/cookbook/creating-webhooks.md
boutell: Yes.

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,93 @@
...
+        // data sent by the afterPublish event includes the document data 
+        // and whether it is being published for the first time
+        // https://v3.docs.apostrophecms.org/reference/server-events.html#afterpublish
+        if (!data.firstTime) {
+          return;
+        }
+
+        await self.apos.http.post('https://slack-webhook-url-here', {
+          body: {
+            text: 'do something cool with the data'
```
boutell, line 25: How about we show how to send the title of the document as part of this? Seems not difficult to do

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,93 @@
...
+
+```javascript
+options: {
+  csrfExceptions: [ '/webhooks' ]
+},
+apiRoutes(self) {
+  return {
+    post: {
+      // change function name to match the desired endpoint URL
+      '/webhooks': async function(req) {
```
boutell, line 57: As discussed I'm fine with this being a rooted URL.

boutell: I'm twitching to ask that this be a real and specific integration with a known service, but despite seeing docs for outgoing webhooks in a zillion services, none are really a great match. So I think it's fine as it is.

boutell (approved):

docs/cookbook/creating-webhooks.md
BoDonkey: Yeah, sure.
I also wasn't sure about not setting a `const` equal to the response that is returned from the `POST`. Do you think people are going to be asking the followup about knowing how they got a `400`?

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,94 @@
...
+        // data sent by the afterPublish event includes the document data 
+        // and whether it is being published for the first time
+        // https://v3.docs.apostrophecms.org/reference/server-events.html#afterpublish
+        if (!data.firstTime) {
+          return;
+        }
+
+        await self.apos.http.post('https://slack-webhook-url-here', {
+          body: {
+            docTitle: data.published.title
```
boutell, line 25: ? Slack won't know what to do with that.

Since this is a slack webhook I was suggesting you make a nice `text` property that utilizes the title to say "page so-and-so was published."

boutell (changes requested):
Re: your question about the slack webhook failing, it is worth including a short note that you can catch exceptions from the webhook if you think it's not critical and shouldn't result in an error in apostrophe, or let them throw in order to make it apparent to the user that there is a problem. If the issue is less critical but still noteworthy, you can catch the exception and do `self.apos.notify(req, 'We were unable to send a notification to slack')` or similar.

docs/cookbook/creating-webhooks.md
BoDonkey: Changed.

docs/cookbook/creating-webhooks.md
```diff
@@ -0,0 +1,94 @@
...
+        // data sent by the afterPublish event includes the document data 
+        // and whether it is being published for the first time
+        // https://v3.docs.apostrophecms.org/reference/server-events.html#afterpublish
+        if (!data.firstTime) {
+          return;
+        }
+        const payload = {
+          text: `Article ${data.published.title} was published.`
+        };
+        const jsonPayload = JSON.stringify(payload);
```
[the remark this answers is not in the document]
BoDonkey: Changed

docs/cookbook/creating-webhooks.md
```diff
@@ -24,7 +24,9 @@ handlers(self) {
         };
         const jsonPayload = JSON.stringify(payload);
```
boutell, line 25: This line is still present.

boutell (changes requested):

docs/cookbook/creating-webhooks.md
BoDonkey: Deleted. Apologies.

boutell (approved):

PR #334, opened 2023-05-31
note on where you can use database queries

boutell:
This is intended to cover a confusion our own team has experienced and that community members have as well, in addition to a very non-obvious point that came up yesterday about calling too early in the startup process.

I didn't really feel like there was a great place to link my reference to "api routes" to in the current docs. Did I miss a page?

### DIFF
--- a/docs/guide/database-queries.md
+++ b/docs/guide/database-queries.md
@@ -352,7 +352,7 @@ Using `self.apos.doc.find` has limitations, however. Significantly, pieces will 
 ## Where in my code can I make database queries?
 
 You can make database queries in async components (as shown above), routes
-(such as in an `apiRoutes` function), [server event handlers](./server-events.md),
+(such as in an [`apiRoutes`](../reference/module-api/module-overview.md#apiroutes-self) function), [server event handlers](./server-events.md),
 middleware and in pretty much any async function in your server-side code.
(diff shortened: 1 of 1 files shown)


docs/guide/database-queries.md
```diff
@@ -10,9 +10,9 @@ This API also provides a layer of security. Queries made through Apostrophe (tha
 We use the terms "query" and "query builders" here. For developers with advanced Apostrophe 2 experience, these are generally the same as the A2 concepts of "cursors" and "cursor filters."
 :::
 
-## Initiating the data query
+## Initiating the database query
```
boutell, line 13: For consistency with the rest of the page and the repo.

BoDonkey:
> This is intended to cover a confusion our own team has experienced and that community members have as well, in addition to a very non-obvious point that came up yesterday about calling too early in the startup process.
> 
> I didn't really feel like there was a great place to link my reference to "api routes" to in the current docs. Did I miss a page?

Would `https://v3.docs.apostrophecms.org/reference/module-api/module-overview.html#apiroutes-self` work?

BoDonkey (approved):
I added a potential link that you could add if you think it is appropriate. Otherwise looks good.

boutell:
Thanks, I took your suggestion.

PR #339, opened 2023-06-05
add doc for redirectToFirstLocale i18n option

ETLaurent:
cf https://github.com/apostrophecms/apostrophe/pull/4179/files

### DIFF
--- a/docs/reference/modules/i18n.md
+++ b/docs/reference/modules/i18n.md
@@ -20,7 +20,6 @@ The module makes an instance of the [i18next](https://npmjs.org/package/i18next)
 |---|---|---|
 |`defaultLocale` | String | The locale that will be used by the UI and server rendering if no other is specified. |
 |`locales` | Object | The set of [locales](../glossary.md#locale) to use in the application. |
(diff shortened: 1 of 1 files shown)


docs/reference/modules/i18n.md
```diff
@@ -20,6 +20,7 @@ The module makes an instance of the [i18next](https://npmjs.org/package/i18next)
 |---|---|---|
 |`defaultLocale` | String | The locale that will be used by the UI and server rendering if no other is specified. |
 |`locales` | Object | The set of [locales](../glossary.md#locale) to use in the application. |
+|`redirectToFirstLocale` | Boolean | If enabled, redirects to the first locale configured with the current requested hostname when all of the locales configured with that hostname do have a prefix. If the request does not match any explicit hostname assigned to locales, redirects to the first locale that does not have a configured hostname, if all the locales without a hostname do have a prefix. |
```
ETLaurent, line 23: could use some better phrasing

BoDonkey: How about: "If enabled, and all locales matching the currently requested hostname have a prefix defined, it will redirect to the locale defined first. If the request does not match any explicit hostname assigned to locales, redirects to the first locale that does not have a configured hostname."
I don't think you have to specify "...if all the locales without a hostname do have a prefix." They would be invalid if they didn't have `hostname` or `prefix`, correct?

ETLaurent: You can have one locale with a configured hostname that does not have a prefix, if I am not mistaken.
Still, I like your version better, I'll use it, thanks

BoDonkey (approved):

PR #352, opened 2023-08-04
add missing snippets

haroun:
## Summary

Add missing snippets to the documentation to improve DX.

When trying the user settings for the first time, it was not clear to me what the actual steps were to achieve the UI language switcher.

This will require some additional work and rephrasing, maybe a screenshot

1. Check `doc/reference/modules/settings.md` preview

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/docs/reference/modules/settings.md
+++ b/docs/reference/modules/settings.md
@@ -67,7 +67,7 @@ module.exports = {
 
 ### Requirements
 
-All fields must be part of the `@apostrophecms/user` schema, except `adminLocale` that will be added automatically to
+All fields added to the subforms except `adminLocale` must already exist as default schema fields of the core `@apostrophecms/user` module or be added at project level. 
 `@apostrophecms/user` if the `adminLocales` option is set in the `@apostrophecms/i18n` module.
 
 <AposCodeBlock>
--- a/docs/reference/modules/settings.md
+++ b/docs/reference/modules/settings.md
@@ -35,6 +35,7 @@ module.exports = {
         // This preview will be automatically generated.
         // preview: '{{ firstName }} {{ lastName }}'
       },
+      // The `adminLocale` option must be configured in the `@apostrophecms/i18n` module for this to be allowed
       adminLocale: {
         fields: [ 'adminLocale' ]
       },
--- a/docs/reference/modules/settings.md
+++ b/docs/reference/modules/settings.md
@@ -35,7 +35,7 @@ module.exports = {
         // This preview will be automatically generated.
         // preview: '{{ firstName }} {{ lastName }}'
       },
-      // The `adminLocale` option must be configured in the `@apostrophecms/i18n` module for this to be allowed
+      // The `adminLocales` option must be configured in the `@apostrophecms/i18n` module for this to be allowed
       adminLocale: {
         fields: [ 'adminLocale' ]
       },
@@ -52,6 +52,7 @@ module.exports = {
       },
       preferences: {
         label: 'Preferences',
+        // The `adminLocales` option must be configured in the `@apostrophecms/i18n` module for this to be allowed
         subforms: [ 'adminLocale' ]
       }
     }
--- a/docs/reference/modules/settings.md
+++ b/docs/reference/modules/settings.md
@@ -59,8 +59,101 @@ module.exports = {
 
 ```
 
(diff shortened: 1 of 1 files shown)


docs/reference/modules/settings.md
```diff
@@ -59,11 +59,102 @@ module.exports = {
...
 
-<template v-slot:cation>
+<template v-slot:caption>
   /modules/@apostrophecms/settings/index.js
 </template>
 </AposCodeBlock>
 
+### Requirements
+
+All fields must be part of the `@apostrophecms/user` schema, except `adminLocale` that will be added automatically to
```
BoDonkey, line 69:
```suggestion
All fields added to the subforms must be included as default fields or added to the `@apostrophecms/user` schema, except `adminLocale`. 
```

docs/reference/modules/settings.md
```diff
@@ -59,11 +59,102 @@ module.exports = {
...
+`@apostrophecms/user` if the `adminLocales` option is set in the `@apostrophecms/i18n` module.
```
BoDonkey, line 70:
```suggestion
If the `adminLocales` option is set in the `@apostrophecms/i18n` module, this field will be automatically added to the available subform schema fields and will add a select input for Admin language to the User management modal.
```

docs/reference/modules/settings.md
```diff
@@ -59,11 +59,102 @@ module.exports = {
 
 ```
 
-<template v-slot:cation>
+<template v-slot:caption>
```
BoDonkey, line 62: Doh!

boutell (changes requested):

docs/reference/modules/settings.md
```diff
@@ -59,11 +60,102 @@ module.exports = {
...
+};
+
+```
+
+<template v-slot:caption>
+  /modules/@apostrophecms/user/index.js
+</template>
+</AposCodeBlock>
+
+<AposCodeBlock>
```
BoDonkey, line 112: We could consider making this shorter, in case the user wants to copy-paste there is less to toss if it doesn't pertain to their project.

BoDonkey (changes requested):
Both Tom and I made suggestions, but feel free to edit specific word usage.

docs/reference/modules/settings.md
```diff
@@ -35,6 +35,7 @@ module.exports = {
         // This preview will be automatically generated.
         // preview: '{{ firstName }} {{ lastName }}'
       },
+      // The `adminLocale` option must be configured in the `@apostrophecms/i18n` module for this to be allowed
```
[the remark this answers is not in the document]
BoDonkey: Oops!

haroun:
I did not changed `displayName` to bind to `title` because there is some `previewComponent` logic in `SettingsDisplayNamePreview` I'm not familiar with yet that must be migrated too.

docs/reference/modules/settings.md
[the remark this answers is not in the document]
haroun: there was a typo also, it is `adminLocales` option

boutell:
I see your point. If we don't add `firstName` and `lastName` to `@apostrophecms/user` in this example, then there is no motivation for the custom previewer.

I suggest breaking it into two steps then: (1) here's a swell config for 99% of sites, no custom preview required. (2) here's a use case with firstName and lastName, thus a use for custom preview.

boutell (changes requested):
I see your point. If we don't add firstName and lastName to @apostrophecms/user in this example, then there is no motivation for the custom previewer.

I suggest breaking it into two steps then: (1) here's a swell config for 99% of sites, no custom preview required. Almost everybody: you're done now!

(2) here's a use case with firstName and lastName, thus a use for custom preview.

haroun:
@BoDonkey I've reassigned you the PR to address the change Tom requested but with `displayName` instead

boutell (approved):
Looks good. I still wish we were starting with a config that works out of the box on every site and then separately talking about fields you might add at project level, but that concern can be better addressed in another way by enabling this module in our starter kits, so that people aren't likely to even get here unless they want to tweak it.

PR #402, opened 2024-09-30
Noticed this text is all about ui/src/index.js files, but the examples are ui/public which does not follow the pattern shown

boutell:


BoDonkey (approved):
Good catch

PR #418, opened 2024-12-16
Update of Vite doc

BoDonkey:
Can you take a first pass at this just to see if I'm in the right universe?


BoDonkey:
Might be easiest to clone locally, `npm I && npm run dev`, then navigate to `http://localhost:5173/guide/vite.html`

myovchev (approved):
I think I already reviewed it in this PR - https://github.com/apostrophecms/docs/pull/419
So I'm approving that to not duplicate my review notes (which can be addressed in the above PR).

PR #419, opened 2024-12-17
Break point preview

BoDonkey:
## Summary

Updates the breakpoint preview documentation to be on its own page plus expands the limitations.

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [X] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


docs/guide/breakpoint-preview.md
```diff
@@ -0,0 +1,159 @@
+
+## Device breakpoint preview
+
+Pages can be enabled to emulate devices, like phones and tablets, making it easy to see how your content will look at different screen sizes.
```
boutell, line 4: "Pages can be enabled to..." is awkward. How about "Apostrophe provides a convenient way to emulate devices..."

docs/guide/breakpoint-preview.md
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+};
+```
+<template v-slot:caption>
+modules/@apostrophecms/asset/index.js
+</template>
+</AposCodeBlock>
+
+You can learn about all the properties that can be passed to the `breakpointPreviewMode` option on the [`@apostrophecms/asset` reference page](/reference/modules/asset.md#breakpointPreviewMode). The two most critical are `enable` and `screens`. If you pass `enable: false` it will remove the breakpoint preview icons and menu from the admin-bar. The `screens` property takes an object where each property is a different breakpoint. Adding the `breakpointPreviewMode` at the project level will override any default values, so you need to pass `enable: true` if you wish to add additional brekpoints.
```
boutell, line 81: admin-bar -> admin bar

boutell: fix "brekpoints"

docs/guide/breakpoint-preview.md
```diff
@@ -0,0 +1,159 @@
...
+
+Each of the individual breakpoint properties in the `screen` object take a `label` that will be displayed to the user when they hover over the icon. The `icon` property supplies the icon that will be displayed if the breakpoint is also given a property of `shortcut: true`. This icon name should be either an icon you can [register in this module](/reference/module-api/module-options.md#icon) or an icon that is [already registered](/reference/module-api/module-overview.md#icons). Finally, each breakpoint needs the `width` and `height` of the container that corresponds to the device being emulated.
```
boutell, line 83: ... ", in web pixels"

If they know, they'll know

docs/guide/breakpoint-preview.md
```diff
@@ -0,0 +1,159 @@
...
+
+---
+
+### Limitations and Differences
+While the ApostropheCMS breakpoint Preview effectively converts many media queries to container queries, there are some key limitations and differences to be aware of. These limitations stem from fundamental differences between **media queries** (which target the viewport) and **container queries** (which target individual containers).
```
boutell, line 88: Consistent capitalization

docs/guide/breakpoint-preview.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### 1. **Unsupported Parameters**
+Certain media query conditions are not supported or do not translate well to container queries:
+- **Height-based queries**:
+  - `min-height` and `max-height` conditions cannot be applied to containers. Container queries only consider the container's width.
+  - Example of unsupported query:
+    ```css
+    @media (min-height: 600px) { ... }
+    ```
+- **Orientation queries**:
```
boutell, line 98: Huh I think we can probably fix this one but that's not a PR comment, it's just an observation.

docs/guide/vite.md
```diff
@@ -104,6 +104,209 @@ For debugging production builds, you can enable source maps to see the original
...
+   ```
+
+2. Configure the plugin in your Vite config
+3. Follow the framework's best practices for file organization within your `ui/src/` directory
+
+::: tip
+Remember that while you have full control over the public build configuration, it's best to start minimal and add only what you need to keep your build process efficient.
+:::
+
+## Built-in Aliases
```
boutell, line 175: Make a clear recommendation here, new devs are only going to retain one. `Modules/` is required for bc reasons but it sounds like `@/` is just better, especially for DX. So we should relegate `Modules/` to a second section that clearly indicates it's not as good.

docs/guide/vite.md
```diff
@@ -104,6 +104,209 @@ For debugging production builds, you can enable source maps to see the original
...
+  }
+}
+```
+
+**Note:** If you've changed your project asset namespace from the default, adjust the `baseUrl` and `exclude` paths accordingly. For example, with a namespace of `my-namespace`:
+- `baseUrl` should be `./apos-build/@apostrophecms/vite/my-namespace`
+- `exclude` should include `apos-build/@apostrophecms/vite/my-namespace/dist`
+
+:::warning
+When following imports in your editor (e.g., Ctrl + Click in VSCode), you'll be taken to the `apos-build` directory rather than the original source code. This is because the `apos-build` directory contains a complete copy of your project's source code (including Admin UI) from all modules (local and npm) and serves as Vite's actual source directory for building the project.
```
boutell, line 237: "Do not modify code in this folder as your changes will be lost. Change files in the original folder." Right? Or is it all symlinked? cc @myovchev

docs/guide/vite.md
```diff
@@ -104,6 +104,209 @@ For debugging production builds, you can enable source maps to see the original
...
+1. **Using the Public Folder**
+```css
+/* File location: ./modules/some-module/public/font.ttf */
+@font-face {
+  font-family: MyFont;
+  src: url("/modules/some-module/font.ttf") format("truetype");
+}
+```
+
+2. **Using Source Root Path**
```
boutell, line 297: It surprises me that this is not using `@`. I guess `src` functions like `@` for Sass for whatever reason? Can this be called out clearly if this difference is real and people need to know?

boutell (changes requested):

docs/guide/vite.md
myovchev: It's not symlinked. I did PoC recently with symlinks and while it would mostly work (creates few solvable problems), I don't think it's worth. It doesn't help DX in any way (editors don't handle it in the way I thought they would do) and we still need watchers because not everything should be symlinked.

docs/guide/vite.md
myovchev: CSS paths don't support aliases. Vite is doing all the hard job. It transforms paths that it can recognize and resolve based on a base asset path property that we provide in different environments. I don't think there is a recommended way here. You either use paths to files in your module `public/` folder (legit feature supported on both UI and back-end) or you use paths to files in your `ui/` folder (can be targeted only in the UI). 

`/src/` is the more clean way for front-end. It targets the real path inside of `apos-build` as Vite sees it. Vite then handles it properly and adds it to the manifest, which then we take and copy for asset deployment. 

`/modules/` is our "little cheat" for UI to let the assets in `module-name/public` still work as before. As a side effect we end up with 2 versions of this file, one is deployed as it is `public/.../modules/module-name/file` (because server templates might use it), and the other `public/../assets/file-123abs` (enumerated by the bundler) that is used only by the UI (be it CSS or JS). 

SASS should now understand `@/`  because it supports the Vite aliases out of the box.

myovchev (changes requested):
It looks good to me (added some clarifications as responses to Tom's feedback and questions). 
What I'm missing is this URL - https://docs.apostrophecms.org/guide/vite.html#resolve-alias-errors
It's a URL that will be shown in both Browser and CLI when `Modules/` alias fails to resolve (because we have a control over that scenario). The #resolve-alias-errors section should clarify that:
- You see the error because of `Modules/` alias can't resolve to the source it targets
- Offer (link up section) info about how `Modules/` are resolved to the real source 
- Probably a good idea to state again that "if you import in your `ui/src/` via `@/` alias a source from `ui/apos` this might happen and is the expected behavior (you shouldn't do that).

docs/guide/breakpoint-preview.md
BoDonkey: Changed.

docs/guide/breakpoint-preview.md
BoDonkey: Changed and fixed.

docs/guide/breakpoint-preview.md
BoDonkey: Change, but "web" seemed odd. Am I missing something?

docs/guide/breakpoint-preview.md
BoDonkey: Fixed throughout to be lowercase.

docs/guide/breakpoint-preview.md
BoDonkey: Yeah, I'm not sure why this is a limitation. It isn't a hard fix (I don't think), just a conversion from `min-height` to `height >=`

docs/guide/vite.md
BoDonkey: Reordered and added a sentence.

docs/guide/vite.md
BoDonkey: Sentence added.

docs/guide/vite.md
BoDonkey: Might need help here to clarify (if needed).

BoDonkey:
I don't know How I managed to mix my two documentation PRs. My apologies to you both.

docs/guide/vite.md
myovchev: @BoDonkey TL;DR aliases are not working for CSS properties (like `url()`), absolute paths do and are normalized in production builds. Use `/modules/` if you put your stuff in `module-name/public` and when you want to access those assets from your Nunjucks templates. Use `/src/module-name/my-asset` when you want to access those files only in the front-end UI.

docs/guide/vite.md
```diff
@@ -198,6 +188,19 @@ import utils from '@/some-module/src/lib/utils.js';
 // Actual path: modules/some-module/ui/apos/mixins/SomeMixin.js
 import SomeMixin from '@/some-module/apos/mixins/SomeMixin.js';
 ```
+::: warning
+Important: When using the `@/` alias, you cannot import files from `ui/apos` directories within files located in `ui/src`. This restriction helps maintain proper separation between admin and public code.
+:::
+
+### The `Modules/` Alias
+
+The `Modules/` alias is available for both public and admin UI builds. It allows you to import modules without worrying about relative paths, **but** restricts you to sources inside `ui/src/` directories.
```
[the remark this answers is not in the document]
BoDonkey: Updated to be technically correct and clearer.

PR #527, opened 2026-03-16
Add url module reference

BoDonkey:
## Summary

This PR adds a reference document for the new `@apostrophecms/url` module.

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/docs/reference/modules/url.md
+++ b/docs/reference/modules/url.md
@@ -16,7 +16,7 @@ Options are passed into the url module by creating a `modules/@apostrophecms/url
 
 | Property | Type | Default | Description |
 |---|---|---|---|
-| [`static`](#static) | Boolean | `false` | Enables static-friendly path-based URLs and URL metadata behavior for static site generation |
+| [`static`](#static) | Boolean | `false` | Switches filter and pagination URLs to path-based format, and enables URL metadata collection for static site generation when used with an external frontend such as Astro |
 
 ### `static`
 
@@ -27,7 +27,7 @@ Examples:
 - Query style (default): `/articles?category=tech&page=2`
 - Static style: `/articles/category/tech/page/2`
 
-This option also enables URL metadata collection used by static frontends to discover all pages that need to be generated, and activates the `GET /api/v1/@apostrophecms/url` endpoint consumed by the `apostrophe-astro` package during `astro build`.
+This option also enables URL metadata collection used by static frontends to discover all pages that need to be generated, and activates the `GET /api/v1/@apostrophecms/url` endpoint consumed by the `apostrophe-astro` package during `astro build`. Performance-intensive operations such as locale URL collection only run when a static build request is actually detected, not on every request.
 
 ::: warning
 `static: true` affects filter and pagination URL format for all frontends connected to the same backend. If you are running both an SSR editorial frontend and a static production build against the same ApostropheCMS instance, make sure your SSR templates use helpers that produce URLs compatible with either format.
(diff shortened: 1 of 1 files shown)


docs/reference/modules/url.md
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+});
+```
+  <template v-slot:caption>
+    app.js
+  </template>
+</AposCodeBlock>
+
+Regardless of which approach you use to set `static: true`, you should also set `staticBaseUrl` so that static URL generation has a known public origin to work from. It can be supplied via the `APOS_STATIC_BASE_URL` environment variable or at the top-level in the `app.js` file. The hard coded value can be used as a convenient fallback for local development and will be overridden by the variable during production deployment.
```
myovchev: This is the recommendation only for Astro static builds, external frontend configurations. The standard Apostrophe projects don't need that. 

The Experimental feature can be used as standalone, not tied to static builds. Internal information. The performance heavy parts (e.g. the locale URLs for a given page will not be "static" but still `/api/v1/xxx` redirects) should be excluded if no static build is actually detected.

docs/reference/modules/url.md
```diff
@@ -0,0 +1,310 @@
...
+            sitemap: false
+          });
+        }
+      }
+    };
+  }
+};
+```
+  <template v-slot:caption>
+    modules/my-module/index.js
```
myovchev: `my-module` should not be used as it has a special internal meaning (`moog`).

docs/reference/modules/url.md
```diff
@@ -0,0 +1,310 @@
...
+
+This endpoint is consumed internally by the `apostrophe-astro` package and is not typically called directly in project-level code.
+
+## Featured methods
+
+The following methods belong to this module and may be useful in project-level code. See the [source code](https://github.com/apostrophecms/apostrophe/tree/main/packages/apostrophe/modules/%40apostrophecms/url) for all methods that belong to this module.
+
+Because this module has an alias, you can call these from another module using the alias path. For example, `self.apos.url.build()`.
+
+### `build(url, [path], data, [...])`
```
myovchev, line 184: A note - I don't think we made `static url` aware. It helps building query string that is not a good thing for static builds. However even in Nunjucks it won't help produce static URLs automatically when `static: true` in  any way.

docs/reference/modules/url.md
```diff
@@ -0,0 +1,310 @@
...
+1. If a hostname is configured for the active locale, `<protocol>://<hostname>` is returned. The global prefix is not appended in this case.
+2. If the request is a static build request, `staticBaseUrl` + prefix is returned.
+3. Otherwise, `apos.baseUrl` + prefix is returned.
+
+`options` accepts:
+
+| Property | Default | Description |
+|---|---|---|
+| `strict` | `false` | When `true` and in a static build context where `staticBaseUrl` is not set, falls back to `apos.baseUrl` rather than returning an empty string. Use when an absolute URL is required, such as for sitemap `<loc>` values |
+| `prefix` | `true` | When `true`, appends `apos.prefix` to the returned URL. Pass `false` to obtain only the origin without the prefix |
```
myovchev, line 236: I just added a new option `relative: false` (by default) so you can have relative (prefix only, no origin) path.
Done in https://github.com/apostrophecms/apostrophe/pull/5356
The code documentation of the method was updated. 
This will become live in the next release!

myovchev (changes requested):

docs/reference/modules/url.md
BoDonkey: @myovchev is the gating behavior (heavy ops only run during actual static builds) something we should document, or treat as an implementation detail? Something like "performance-intensive URL collection only runs when a static build is detected"

docs/reference/modules/url.md
BoDonkey: Changed.

docs/reference/modules/url.md
BoDonkey: Added a clarifying note.

docs/reference/modules/url.md
BoDonkey: Added

docs/reference/modules/url.md
myovchev: It's more about do we degrade standard app performance while `static: true` URLs are enabled. The example I gave is **static-build** vs static URLs. In static build we really want ALL URLs static. When running the standard app, we prioritize performance. I think this is the important bit, it should be already clear that the meta enumeration process is heavy and happens only when you do static building.

myovchev (approved):

PR #528, opened 2026-03-17
document the image and file modules, and the pretty URLs options specifically

boutell:


BoDonkey:
For the file.md documentation, there is some text that I would have put together a little differently, but these are "me" stylistic issues, not things that really need changing. I would have been slightly more formal in tone with the "Interesting properties" -> "Selected properties", or... I would likely have added the warning about pretttyUrls performance into an info box of some type. But those are just my style, nothing that needs changing.

docs/reference/modules/image.md
```diff
@@ -0,0 +1,254 @@
...
+| `creditUrl` | URL | A URL associated with the image credit. |
+| `_tags` | Relationship | Tags for organizing images, related to `@apostrophecms/image-tag` pieces. |
+
+## Template helpers
+
+Template helpers are methods available for use in template files. Because this module has an alias, you can call these in templates using the alias path, e.g., `apos.image.first(within)`.
+
+### `first(within, options)`
+
+A convenience wrapper for [`apos.attachment.first()`](/reference/modules/attachment.md#first-within-options) that automatically filters for image attachments only (equivalent to passing `{ group: 'images' }` in the options). Returns the first image attachment found within the given document or area, or `undefined` if none is found.
```
BoDonkey: Looking at the module code, I think that if no image is found, it returns `null`:
```
first(within, options) {
        if (!within) {
          return null;
        }
        return self.first(within, options);
      },
```

docs/reference/modules/image.md
```diff
@@ -0,0 +1,254 @@
...
+{% endfor %}
+```
+<template v-slot:caption>
+views/show.html
+</template>
+</AposCodeBlock>
+
+### `srcset(attachment, cropFields)`
+
+Generates an HTML `srcset` attribute value for responsive images. This produces a comma-separated list of URLs at each configured image size along with their width descriptors, suitable for use in an `<img>` tag's `srcset` attribute.
```
BoDonkey, line 177: Should we mention that SVG aren't sized so it will return an empty string (like the caveat with `isCroppable`)

docs/reference/modules/image.md
```diff
@@ -0,0 +1,254 @@
...
+extends: '@apostrophecms/piece-type'
+---
+
+# `@apostrophecms/image`
+
+**Alias:** `apos.image`
+
+<AposRefExtends :module="$frontmatter.extends" />
+
+This module manages the image library in Apostrophe. Images are [piece-type](/guide/pieces.md) documents with an attachment field restricted to the `images` file group. The module provides template helpers for retrieving images and generating `srcset` attributes for responsive images.
```
BoDonkey, line 11: Since the file module reference mentions the `file-widget`, do you think it is worth mentioning here, potentially with a side-note that it adds images through a relationship?

BoDonkey:
Side note unrelated to this PR - for the `@apostrophecms/image/index.js` file - in the comment block at the top:
```
// Together with
// [@apostrophecms/image-widget](../@apostrophecms/image-widget/index.html),
// this module provides a simple way to add downloadable PDFs and the like to a
// website, and to manage a library of them for reuse.
```
Ermmmm - this is just a function of the file module, right?

BoDonkey (changes requested):
Mostly for the 'undefined' vs 'null' return in the image documentation.

docs/reference/modules/image.md
```diff
@@ -176,6 +178,8 @@ Generates an HTML `srcset` attribute value for responsive images. This produces
 
 In most cases you do not need to pass `cropFields`. When images are used via `@apostrophecms/image-widget`, the cropping coordinates are stored in the widget's relationship fields and applied automatically. The `cropFields` parameter is available for advanced cases where you need to manually specify crop dimensions (an object with `top`, `left`, `width`, and `height` properties).
 
+Note that `svg` files are not sized so this function returns an empty string. It is important to offern the `src` attribute also for that reason.
```
BoDonkey: "offern"->"offer"

BoDonkey (changes requested):

BoDonkey (approved):

PR #532, opened 2026-04-16
Postgres

boutell:
Initial documentation for the new Postgres branch.

cc @myovchev for easier evaluation

I'm operating on the assumption that this will ship before a new CLI so it's written from that perspective but obviously I am enthusiastic about the goal of a CLI that says, "who wants a slice of cake? I've got everything you could want..."


BoDonkey (changes requested):
I think the "Multi-tenant" section needs to be clearer. With respect to SQLite, the current documentation seems to call out PostgreSQL as a special case, without making clear that, in general, the environment variable needs to be changed. With respect to the database name, the need to add any site prefix to the database name is important. I think the clarification that SQLite needs to use `DB_URI` will also clarify that the database name should contain `-dashboard`.
I know all this information is surfaced in the multisite readme, but I think since you started down the path, it needs to be complete. Alternatively, you could truncate the multisite section to acknowledge that the new database options exist and that users can refer to the multisite documentation for further info.

BoDonkey (approved):

PR #550, opened 2026-06-24
Adds `tagPickerPerPage`

BoDonkey:
## Please indicate which branch this PR should merge into:
*Check one*

- [ X ] main
- [ ] latest
- [ ] stable

- [ ] Check if this PR will be resubmitted against another branch
## Summary

Adds the new `tagPickerPerPage` option documentation to the image module reference.DO NOT MERGE UNTIL RELEASE.

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/docs/reference/modules/image.md
+++ b/docs/reference/modules/image.md
@@ -20,6 +20,7 @@ This module is often used together with the [`@apostrophecms/image-widget`](/ref
 |---|---|---|
 | [`insertViaUpload`](#insertviaupload) | Boolean | Defaults to `true`. Enables uploading images directly from the manager modal. |
 | [`perPage`](#perpage) | Integer | Defaults to `50`. The number of images fetched per batch via infinite scroll in the manager modal. |
+| `tagPickerPerPage` | Integer | Defaults to `400`. The number of tags to load into the tag popover at once. Increase for sites with a large number of tags. |
 | [`searchable`](#searchable) | Boolean | Defaults to `false`. Images are excluded from search results by default. |
 | [`slugPrefix`](#slugprefix) | String | Defaults to `'image-'`. Prefix applied to image slugs. |
 | [`sort`](#sort) | Object | Defaults to `{ createdAt: -1 }`. Sort order for the image library. |


myovchev (approved):

docs/reference/modules/image.md
```diff
@@ -20,6 +20,7 @@ This module is often used together with the [`@apostrophecms/image-widget`](/ref
 |---|---|---|
 | [`insertViaUpload`](#insertviaupload) | Boolean | Defaults to `true`. Enables uploading images directly from the manager modal. |
 | [`perPage`](#perpage) | Integer | Defaults to `50`. The number of images fetched per batch via infinite scroll in the manager modal. |
+| `tagPickerPerPage` | Integer | Defaults to `400`. The number of tags to load into the tag popover at once. Increase for sites with a large number of tags. |
```
boutell, line 23: Wait... if our advice is to just keep increasing it... are we saying the feature is effectively broken unless we just switch to fetching all the tags all the time? Because if that is the case we should just provide a way to get 100% of the tags each time unless and until we have the energy to do something more efficient that isn't broken.

boutell (changes requested):

docs/reference/modules/image.md
myovchev: It should load everything, yes. The batch popup don't support pagination. However, not so bad, because we now eliminate duplicates on backend. So if you don't find it and create it, the existing will be used. 
So no negatives ATM beside the obvious UX "tag is not found" problem.

boutell: I think we should fix the bug in the simplest way (by loading everything, repeated calls until no more pages) not document the bug as something the user must work around.

boutell: The option is problematic, it has no good outcomes for values other than infinity

boutell: Is there any frontend code that actually tries to use this option, I can only trace it as far as getBrowserData

myovchev: I agree. Now that I look at it, there is a better solution. 

A possible solution:
- a `prob` request for one item so we can get the `total` records
- a real request that requests the total as per page value

A minimal overhead, no loops. And in fact no complexity. 
Captured in PRO-9642 and it will be done today. 

@BoDonkey You can close this PR.

